PR #12, opened 2017-09-04
Committable Source

alexvaluyskiy:
The first version of CommitableSource. Fixed https://github.com/alexvaluyskiy/Akka.Streams.Kafka/issues/2

Batching is not implemented in this pull request

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,24 @@ Consumer.PlainSource(consumerSettings, subscription)
         Console.WriteLine($"Consumer: {result.Topic}/{result.Partition} {result.Offset}: {result.Value}");
     }, materializer);
 ```
+
+### Committable Consumer
+The `Consumer.CommittableSource` makes it possible to commit offset positions to Kafka.
+
+Compared to auto-commit this gives exact control of when a message is considered consumed.
+
+If you need to store offsets in anything other than Kafka, `PlainSource` should be used instead of this API.
+
+This is useful when “at-least once delivery” is desired, as each message will likely be delivered one time but in failure cases could be duplicated.
+
+```C#
+Consumer.CommitableSource(consumerSettings, Subscriptions.Topics("topic1"))
+    .SelectAsync(1, elem =>
+    {
+        return elem.CommitableOffset.Commit();
+    })
+    .RunWith(Sink.Ignore<CommittedOffsets>(), _materializer);
+```
+The above example uses separate mapAsync stages for processing and committing. This guarantees that for parallelism higher than 1 we will keep correct ordering of messages sent for commit.
+
+Committing the offset for each message as illustrated above is rather slow. It is recommended to batch the commits for better throughput, with the trade-off that more messages may be re-delivered in case of failures.
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -56,7 +56,41 @@ namespace Akka.Streams.Kafka.Tests.Integration
         }
 
         [Fact]
-        public async Task PlainSource_consumes_messages_from_KafkaProducer_with_topicPartition_assignment()
+        public async Task CommitableSource_consumes_messages_from_Producer_without_commits()
+        {
+            int elementsCount = 100;
+            var topic1 = CreateTopic(1);
+            var group1 = CreateGroup(1);
+
+            await GivenInitializedTopic(topic1);
+
+            await Source
+                .From(Enumerable.Range(1, elementsCount))
+                .Select(elem => new ProduceRecord<Null, string>(topic1, null, elem.ToString()))
+                .Via(Dsl.Producer.CreateFlow(ProducerSettings))
+                .RunWith(Sink.Ignore<Message<Null, string>>(), _materializer);
+
+            var consumerSettings = CreateConsumerSettings(group1);
+
+            var probe = Dsl.Consumer
+                .CommitableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
+                .Where(c => !c.Record.Value.Equals(InitialMsg))
+                .Select(c =>
+                {
+                    Output.WriteLine($"Consumed: {c.Record.Value}");
+                    return c.Record.Value;
+                })
+                .RunWith(this.SinkProbe<string>(), _materializer);
+
+            probe
+                .Request(elementsCount)
+                .ExpectNextN(Enumerable.Range(1, elementsCount).Select(c => c.ToString()));
+
+            probe.Cancel();
+        }
+
+        [Fact]
+        public async Task CommitableSource_consumes_messages_from_Producer_with_commit_on_each_element()
         {
             var topic1 = CreateTopic(1);
             var group1 = CreateGroup(1);
@@ -67,18 +101,19 @@ namespace Akka.Streams.Kafka.Tests.Integration
                 .From(Enumerable.Range(1, 100))
                 .Select(elem => new ProduceRecord<Null, string>(topic1, null, elem.ToString()))
                 .Via(Dsl.Producer.CreateFlow(ProducerSettings))
+                .Select(record => record)
                 .RunWith(Sink.Ignore<Message<Null, string>>(), _materializer);
 
             var consumerSettings = CreateConsumerSettings(group1);
-
             var committedElements = new ConcurrentQueue<string>();
 
             var (_, probe1) = Dsl.Consumer.CommitableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
-                //.WhereNot(c => c.Record.Value == InitialMsg)
+                .WhereNot(c => c.Record.Value == InitialMsg)
                 .SelectAsync(10, elem =>
                 {
                     return elem.CommitableOffset.Commit().ContinueWith(t =>
                     {
+                        Output.WriteLine($"Consumed: {elem.Record.Value}");
                         committedElements.Enqueue(elem.Record.Value);
                         return Done.Instance;
                     });
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -1,0 +1,99 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Akka.Configuration;
+using Akka.Streams.Dsl;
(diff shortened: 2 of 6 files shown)


src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            await Source
+                .From(Enumerable.Range(1, 100))
+                .Select(elem => new ProduceRecord<Null, string>(topic1, null, elem.ToString()))
+                .RunWith(Dsl.Producer.PlainSink(ProducerSettings), _materializer);
+
+            var consumerSettings = CreateConsumerSettings(group1);
+            var committedElements = new ConcurrentQueue<string>();
+
+            var (_, probe1) = Dsl.Consumer.CommitableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
```
alexvaluyskiy: It should be `Subscription.Topics('topic1')` but it is too slow

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            probe1.Request(25);
+
+            foreach (var _ in Enumerable.Range(1, 25))
+            {
+                probe1.ExpectNext(Done.Instance, TimeSpan.FromSeconds(10));
+            }
+                
+            probe1.Cancel();
+
+            // Await.result(control.isShutdown, remainingOrDefault)
```
alexvaluyskiy: Control is not implemented at the moment

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
```diff
@@ -0,0 +1,179 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Streams.Kafka.Messages;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.Stage;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages
+{
+    internal class CommitableConsumerStage<K, V, Msg> : GraphStageWithMaterializedValue<SourceShape<Msg>, Task>
```
alexvaluyskiy: This class is similar to `PlainSource`, but I don;t want to merge them until API will be finished

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        //
+        // Async callbacks
+        //
+
+        private void MessagesReceived(Message<K, V> message)
+        {
+            var consumer = _consumer;
+            var commitableOffset = new CommitableOffset(
+                () => consumer.CommitAsync(),
```
alexvaluyskiy: Is it fine to pass Task here? Is it thread-safe?

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
```diff
@@ -0,0 +1,178 @@
...
+                new PartitionOffset("groupId", message.Topic, message.Partition, message.Offset));
```
alexvaluyskiy: I should pass groupID here

PartitionOffset could be redundant. I could pass `Message<K, V>` instead

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
Silv3rcircl3: In terms of Akka it is since the task doesn't touch anything from the stage put I don't know what happens with the consumer if you poll/commit from different threads. Also why passing a func around ? Wouldn't it be easier to pass the consumer directly into the CommittableOffsset class and call _consumer.CommitAsync ? Another question, what happens if one calls Commit multiple times on the same message ? Should it be possible ?

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
Silv3rcircl3: Why do your stages return a finished Task as materialized value?

README.md
```diff
@@ -78,4 +78,25 @@ Consumer.PlainSource(consumerSettings, subscription)
...
+
+```C#
+Consumer.CommitableSource(consumerSettings, Subscriptions.Topics("topic1"))
+    .SelectAsync(1, elem =>
+    {
+        return elem.CommitableOffset.Commit();
+    })
+    .RunWith(Sink.Ignore<CommittedOffsets>(), _materializer);
+```
+The above example uses separate mapAsync stages for processing and committing. This guarantees that for parallelism higher than 1 we will keep correct ordering of messages sent for commit.
```
Silv3rcircl3: SelectAsync instead of mapAsync

README.md
```diff
@@ -78,4 +78,25 @@ Consumer.PlainSource(consumerSettings, subscription)
...
+The `Consumer.CommittableSource` makes it possible to commit offset positions to Kafka.
+
+Compared to auto-commit this gives exact control of when a message is considered consumed.
+
+If you need to store offsets in anything other than Kafka, `PlainSource` should be used instead of this API.
+
+This is useful when “at-least once delivery” is desired, as each message will likely be delivered one time but in failure cases could be duplicated.
+
+```C#
+Consumer.CommitableSource(consumerSettings, Subscriptions.Topics("topic1"))
```
Horusiath: I guess it should be `Subscriptions.Topic` not `Topic` (since we're picking a single topic, plural form doesn't sound right).

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
Horusiath: My idea was to make materialized value implement a `IDisposable` object - this way it can be set inside using block, so we don't have to worry about not-disposed connection in case of errors.

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var producer = ProducerSettings.CreateKafkaProducer();
+            await producer.ProduceAsync(topic, null, InitialMsg, 0);
+            producer.Dispose();
+        }
+
+        private ConsumerSettings<Null, string> CreateConsumerSettings(string group)
+        {
+            return ConsumerSettings<Null, string>.Create(Sys, null, new StringDeserializer(Encoding.UTF8))
+                .WithBootstrapServers(KafkaUrl)
+                .WithProperty("auto.offset.reset", "earliest")
```
Horusiath: This doesn't have to happen now, but my idea was to extract producer/consumer config values from documentation and expose them as `Property<T>(string config, T value)`. This way you can have type-safe properties, that could be suggested by intellisense.

README.md
alexvaluyskiy: It could accept one or many topics `Topics(params string[] topics)`

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
alexvaluyskiy: Or we could use `CancellationTokenSource` here

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
alexvaluyskiy: It is a temporary solution. I should fix it

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
```diff
@@ -0,0 +1,178 @@
...
+            _subscription = subscription;
+            _out = shape.Outlets.FirstOrDefault();
+
+            SetHandler(_out, onPull: () =>
+            {
+                if (_buffer.Count > 0)
+                {
+                    Push(_out, _buffer.Dequeue());
+                }
+                else
```
Silv3rcircl3: You should make this another if and check if buffer is empty, i.e. If the last element was send downstream immediately poll and don't wait for new demand .

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
```diff
@@ -0,0 +1,178 @@
...
+        private void HandleOnMessage(object sender, Message<K, V> message) => _messagesReceived.Invoke(message);
+
+        // TODO: how I should react?
+        private void HandleConsumeError(object sender, Message message) { }
+
+        private void HandleOnError(object sender, Error error)
+        {
+            Log.Error(error.Reason);
+
+            if (!KafkaExtensions.IsBrokerErrorRetriable(error) && !KafkaExtensions.IsLocalErrorRetriable(error))
```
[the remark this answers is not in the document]
alexvaluyskiy: Confluent.Kafka does not produce any exceptions, only error codes

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
alexvaluyskiy: @Silv3rcircl3 is it possible to create an integration test for this case?

Silv3rcircl3: You could put two messages into Kafka, request 1, wait for the timer Intervall, i.e. Buffer = 1 and therefore timer isn't scheduled again, put another message into Kafka, request another one ( buffer is not empty therefore onPull will currently only push without polling Kafka ), delete message from Kafka, request another one, buffer is empty -> tries to pull but Kafka is empty too so test will fail. Test should pass if you check onPull if buffer is empty after pushing an element.

src/Akka.Streams.Kafka/Stages/CommitableConsumerStage.cs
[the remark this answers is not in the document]
Silv3rcircl3: Than I would maybe use a custom Excpetion that contains the error code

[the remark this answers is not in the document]
alexvaluyskiy: Confluent.Kafka has `KafkaExteption`, but the driver itself does not use it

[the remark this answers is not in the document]
Silv3rcircl3: But currently User can't decide what he want to do in case of failure, I.e. If one set Decider to Resume "globally" the streams should never fail.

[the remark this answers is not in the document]
alexvaluyskiy: Got it, I will fix it

PR #14, opened 2017-09-06
Use deciders for exception handling

alexvaluyskiy:
Confluent.Kafka does not produce any exception, so I'm using only two my exceptions
- SerializationException - only in `OnConsumeError` event handler. `Producer` does not react on serialization error
- KafkaException - both in producer and consumer - for non-retriable errors

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
@@ -1,16 +1,19 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Serialization;
 using System.Text;
 using System.Threading.Tasks;
 using Akka.Configuration;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Messages;
 using Akka.Streams.Kafka.Settings;
+using Akka.Streams.Supervision;
 using Akka.Streams.TestKit;
 using Confluent.Kafka;
 using Confluent.Kafka.Serialization;
 using FluentAssertions;
+using MessagePack.Resolvers;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -24,8 +27,15 @@ namespace Akka.Streams.Kafka.Tests.Integration
 
         private readonly ActorMaterializer _materializer;
 
+        public static Config Default()
+        {
+            return ConfigurationFactory.ParseString("akka.loglevel = DEBUG")
+                .WithFallback(ConfigurationFactory.FromResource<ConsumerSettings<object, object>>(
+                        "Akka.Streams.Kafka.reference.conf"));
+        }
+
         public PlainSourceIntegrationTests(ITestOutputHelper output) 
-            : base(ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf"), null, output)
+            : base(Default(), nameof(PlainSourceIntegrationTests), output)
         {
             _materializer = Sys.Materializer();
         }
@@ -43,6 +53,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
         {
             var producer = ProducerSettings.CreateKafkaProducer();
             await producer.ProduceAsync(topic, null, InitialMsg, 0);
+            producer.Flush(TimeSpan.FromSeconds(1));
             producer.Dispose();
         }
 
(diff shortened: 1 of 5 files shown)


src/Akka.Streams.Kafka/Stages/ProducerStage.cs
```diff
@@ -97,7 +105,21 @@ private void OnProducerError(object sender, Error error)
...
+                var exception = new KafkaException(error);
+                switch (_decider(exception))
+                {
+                    case Directive.Stop:
+                        // Throw
+                        _completionState.TrySetException(exception);
+                        FailStage(exception);
+                        break;
+                    case Directive.Resume:
+                        // TODO: resume the stream
```
alexvaluyskiy: I don't know what to do here. Kafka driver itself has a retry policy for message delivery. If I get unretrievable error, that's mean I should not try again.

src/Akka.Streams.Kafka/Stages/ProducerStage.cs
```diff
@@ -97,7 +105,21 @@ private void OnProducerError(object sender, Error error)
...
+                        break;
+                    case Directive.Restart:
+                        // TODO: restart the stream
```
alexvaluyskiy: What should I do here? Azure implementation try to take the next element or complete the stage

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -125,7 +150,21 @@ private void HandleOnError(object sender, Error error)
...
+                var exception = new KafkaException(error);
+                switch (_decider(exception))
+                {
+                    case Directive.Stop:
+                        // Throw
+                        _completion.TrySetException(exception);
+                        FailStage(exception);
+                        break;
+                    case Directive.Resume:
+                        // TODO: resume the stream
```
alexvaluyskiy: @Silv3rcircl3 help me here

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -125,7 +150,21 @@ private void HandleOnError(object sender, Error error)
...
+                        break;
+                    case Directive.Restart:
+                        // TODO: restart the stream
```
alexvaluyskiy: @Silv3rcircl3 help me here

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -117,6 +127,21 @@ public override void PostStop()
...
+            var exception = new SerializationException(message.Error.Reason);
+            switch (_decider(exception))
+            {
+                case Directive.Stop:
+                    // Throw
+                    _completion.TrySetException(exception);
+                    FailStage(exception);
+                    break;
+                case Directive.Resume:
+                    // TODO: resume the stream
```
alexvaluyskiy: @Silv3rcircl3 help me here

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -117,6 +127,21 @@ public override void PostStop()
...
+                    break;
+                case Directive.Restart:
+                    // TODO: restart the stream
```
alexvaluyskiy: @Silv3rcircl3 help me here

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -5,31 +5,35 @@
...
+            Subscription = subscription;
             Shape = new SourceShape<Msg>(Out);
+            Settings = settings;
+            Subscription = subscription;
         }
 
         public override ILogicAndMaterializedValue<Task> CreateLogicAndMaterializedValue(Attributes inheritedAttributes)
         {
-            return new LogicAndMaterializedValue<Task>(new KafkaSourceStage<K, V>(_settings, _subscription, Shape), Task.CompletedTask);
+            var completion = new TaskCompletionSource<NotUsed>();
```
alexvaluyskiy: @Silv3rcircl3 I've added a TaskCompletionSource, as you asked

src/Akka.Streams.Kafka/Stages/ProducerStage.cs
Horusiath: What about resetting connection (disposing old one, setting a new)?

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
Silv3rcircl3: Don't need to do anything here, just keep going.

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
Silv3rcircl3: Restart is mostly used in places where you have some initial state like StatefulSelect and on restart current state is replced with the initial one. I'm not sure if it makes sense here, if not you can just check for stop and otherwise ignore the excpetion.

Silv3rcircl3: Also you maybe need to call Poll from here if Kafka isn't sending you new messages otherwise.

alexvaluyskiy:
@Silv3rcircl3 @Horusiath  I've changed the scope of this task. Right now I'm using `Decider` only for deserialization errors in a consumer. Can't do the same for Producer, because its API is limited.

Kafka driver has its own retry logic, don't want to duplicate it

src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
```diff
@@ -151,7 +162,62 @@ private TestSubscriber.Probe<string> CreateProbe(ConsumerSettings<Null, string>
...
+
+            var probe = Dsl.Consumer
+                .PlainSource(settings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
+                .WithAttributes(ActorAttributes.CreateSupervisionStrategy(Deciders.StoppingDecider))
+                .Select(c => c.Value)
+                .RunWith(this.SinkProbe<int>(), _materializer);
+
+            var error = probe.Request(elementsCount).ExpectEvent(TimeSpan.FromSeconds(10));
+            error.Should().BeOfType<TestSubscriber.OnError>();
+            ((TestSubscriber.OnError)error).Cause.Should().BeOfType<SerializationException>();
```
alexvaluyskiy: Streams TestKit has limited support for timeouts, so I can use only `ExpectEvent` and `ExpectNext`

PR #15, opened 2017-09-07
Added Pause/Resume

alexvaluyskiy:


src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -102,6 +111,7 @@ public override void PreStart()
             _messagesReceived = GetAsyncCallback<Message<K, V>>(MessagesReceived);
             _partitionsAssigned = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsAssigned);
             _partitionsRevoked = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsRevoked);
+            ScheduleRepeatedly(TimerKey, _settings.PollInterval);
```
alexvaluyskiy: We can't use `ScheduleOnce` with Confluent.Kafka, because if we stop polling the Kafka we just drop a connection, and we won't get any errors. We should poll it without any pauses

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -182,21 +192,26 @@ private void PartitionsAssigned(IEnumerable<TopicPartition> partitions)
...
         {
-            // TODO: should I call `Poll` if there are no assignments? Like in `Subscribe` flow
             _consumer.Poll(_settings.PollTimeout);
 
-            if (_buffer.Count == 0)
-                ScheduleOnce(TimerKey, _settings.PollInterval);
+            if (!isPaused && _buffer.Count > _settings.BufferSize)
+            {
+                Log.Debug($"Polling paused, buffer is full");
+                _consumer.Pause(assignedPartitions);
```
alexvaluyskiy: Is the buffer is full - just stop receiving messages from specified topics

src/Akka.Streams.Kafka/Stages/ConsumerStage.cs
```diff
@@ -68,6 +71,12 @@ public KafkaSourceStageLogic(KafkaSourceStage<K, V, Msg> stage, Attributes attri
                 }
                 else
                 {
+                    if (isPaused)
```
alexvaluyskiy: If a downstream ask for an item and we were paused and have an empty buffer - just resume consuming messages from Kafka

PR #30, opened 2019-08-09
Update Confluence.Kafka to 1.1.0

IgorFedchenko:
The goal of this PR is to update Confluence.Kafka package version to latest (1.1.0).

While trying to keep PR as small as possible, big part of api, exposed by kafka, has been changed since old version. I did not touch current implementation design yet (so the buffering in consumer still exists), but still had to update code pretty much to make it compile and pass tests.

So, here is a list of changes:

1. There is no StringDeserializer anymore, and it is not needed because utf8 serializer used by default
2. Also there is no more MessageAndMeta class to wrap both topic info and message data. But it is useful to have stream of MessageAndMeta passed to producer, so I implemented it as a part of internal models.
3. ConsumeResult is our new ConsumerRecord (they just renamed it)
4. There are no events for incoming messages or partition assignments anymore - all handling should be configured using builder pattern when creating consumer/producer. So I removed all subscriptions, and passed handlers to consumer/producer creation factory:

 ```csharp
public Confluent.Kafka.IConsumer<TKey, TValue> CreateKafkaConsumer(Action<IConsumer<TKey, TValue>, Error> consumeErrorHandler = null,
                                                                   Action<IConsumer<TKey, TValue>, List<TopicPartition>> partitionAssignedHandler = null,
                                                                   Action<IConsumer<TKey, TValue>, List<TopicPartitionOffset>> partitionRevokedHandler = null)
{ 
    return new Confluent.Kafka.ConsumerBuilder<TKey, TValue>(this.Properties)
        .SetKeyDeserializer(this.KeyDeserializer)
        .SetValueDeserializer(this.ValueDeserializer)
        .SetErrorHandler((c, e) => consumeErrorHandler?.Invoke(c, e))
        .SetPartitionsAssignedHandler((c, partitions) => partitionAssignedHandler?.Invoke(c, partitions))
        .SetPartitionsRevokedHandler((c, partitions) => partitionRevokedHandler?.Invoke(c, partitions))
        .Build();
}
 ```

5. Pull is substituted by Consume, and the error handling was changed in this version - there is no OnError event anymore, we have to catch exceptions in place, like this:

 ```csharp
try
{
    var message = _consumer.Consume(_settings.PollTimeout);
    if (message == null)
        return;

    HandleMessage(message);
}
catch (ConsumeException ex)
{
    HandleError(ex.Error);
}
 ```

6. Most of the changes were make in consumer stages - and there is a lot of duplication for PlainSource and Commitable source - did not touch that yet, so you can see same changes applied twice to different classes.

This is my first PR, so any comments of how to make future PR's better are welcome!

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -6,10 +6,10 @@ using System.Threading.Tasks;
 using Akka.Configuration;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Dsl;
+using Akka.Streams.Kafka.Messages;
 using Akka.Streams.Kafka.Settings;
 using Akka.Streams.TestKit;
 using Confluent.Kafka;
-using Confluent.Kafka.Serialization;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -35,7 +35,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
         private string CreateGroup(int number) => $"group-{number}-{Uuid}";
 
         private ProducerSettings<Null, string> ProducerSettings =>
-            ProducerSettings<Null, string>.Create(Sys, null, new StringSerializer(Encoding.UTF8))
+            ProducerSettings<Null, string>.Create(Sys, null, null)
                 .WithBootstrapServers(KafkaUrl);
 
         private async Task GivenInitializedTopic(string topic)
@@ -48,13 +48,13 @@ namespace Akka.Streams.Kafka.Tests.Integration
 
         private ConsumerSettings<Null, string> CreateConsumerSettings(string group)
         {
-            return ConsumerSettings<Null, string>.Create(Sys, null, new StringDeserializer(Encoding.UTF8))
+            return ConsumerSettings<Null, string>.Create(Sys, null, null)
                 .WithBootstrapServers(KafkaUrl)
                 .WithProperty("auto.offset.reset", "earliest")
                 .WithGroupId(group);
         }
 
-        [Fact(Skip = "Needs IMPL")]
+        [Fact/*(Skip = "Needs IMPL")*/]
         public async Task CommitableSource_consumes_messages_from_Producer_without_commits()
         {
             int elementsCount = 100;
@@ -83,7 +83,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
             probe.Cancel();
         }
 
-        [Fact(Skip = "Needs IMPL")]
+        [Fact/*(Skip = "Needs IMPL")*/]
         public async Task CommitableSource_resume_from_commited_offset()
         {
             var topic1 = CreateTopic(1);
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -54,7 +54,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
                 .WithGroupId(group);
         }
 
-        [Fact/*(Skip = "Needs IMPL")*/]
+        [Fact]
         public async Task CommitableSource_consumes_messages_from_Producer_without_commits()
         {
             int elementsCount = 100;
@@ -83,7 +83,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
             probe.Cancel();
         }
 
-        [Fact/*(Skip = "Needs IMPL")*/]
+        [Fact]
         public async Task CommitableSource_resume_from_commited_offset()
         {
             var topic1 = CreateTopic(1);
--- a/src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
+++ b/src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using Akka.Streams.Kafka.Dsl;
 using Confluent.Kafka;
 
@@ -10,13 +11,13 @@ namespace Akka.Streams.Kafka.Messages
     /// </summary>
     public sealed class CommittableMessage<K, V>
     {
-        public CommittableMessage(ConsumerRecord<K, V> record, CommitableOffset commitableOffset)
+        public CommittableMessage(ConsumeResult<K, V> record, CommitableOffset commitableOffset)
         {
             Record = record;
             CommitableOffset = commitableOffset;
         }
 
-        public ConsumerRecord<K, V> Record { get; }
+        public ConsumeResult<K, V> Record { get; }
 
         public CommitableOffset CommitableOffset { get; }
     }
@@ -31,9 +32,9 @@ namespace Akka.Streams.Kafka.Messages
     /// </summary>
     public class CommitableOffset
     {
-        private readonly Func<CommittedOffsets> _task;
+        private readonly Func<List<TopicPartitionOffset>> _task;
 
-        public CommitableOffset(Func<CommittedOffsets> task, PartitionOffset offset)
+        public CommitableOffset(Func<List<TopicPartitionOffset>> task, PartitionOffset offset)
         {
             _task = task;
             Offset = offset;
@@ -41,7 +42,7 @@ namespace Akka.Streams.Kafka.Messages
 
         public PartitionOffset Offset { get; }
 
-        public CommittedOffsets Commit()
+        public List<TopicPartitionOffset> Commit()
         {
             return _task();
         }
--- a/src/Akka.Streams.Kafka/Messages/MessageAndMeta.cs
+++ b/src/Akka.Streams.Kafka/Messages/MessageAndMeta.cs
@@ -1,0 +1,16 @@
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Messages
+{
+    /// <summary>
+    /// Container for storing message and topic/parition info
+    /// </summary>
+    /// <typeparam name="K">Type of key</typeparam>
+    /// <typeparam name="V">Type of value</typeparam>
+    public class MessageAndMeta<K, V>
+    {
+        public Message<K, V> Message { get; set; }
+        public string Topic { get; set; }
+        public TopicPartition TopicPartition { get; set; }
+    }
+}
--- a/src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
+++ b/src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
@@ -1,8 +1,9 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using Akka.Actor;
-using Akka.Configuration;
-using Confluent.Kafka.Serialization;
+using Confluent.Kafka;
+using Config = Akka.Configuration.Config;
 
 namespace Akka.Streams.Kafka.Settings
 {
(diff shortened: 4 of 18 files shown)


src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
```diff
@@ -31,17 +32,17 @@ public CommittableMessage(ConsumerRecord<K, V> record, CommitableOffset commitab
     /// </summary>
     public class CommitableOffset
     {
-        private readonly Func<CommittedOffsets> _task;
+        private readonly Func<List<TopicPartitionOffset>> _task;
```
Aaronontheweb: Why the change?

src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
```diff
@@ -84,7 +85,17 @@ public ConsumerSettings(IDeserializer<TKey> keyDeserializer, IDeserializer<TValu
                 dispatcherId: dispatcherId ?? this.DispatcherId,
                 properties: properties ?? this.Properties);
 
-        public Confluent.Kafka.IConsumer<TKey, TValue> CreateKafkaConsumer() =>
-            new Confluent.Kafka.Consumer<TKey, TValue>(this.Properties, this.KeyDeserializer, this.ValueDeserializer);
+        public Confluent.Kafka.IConsumer<TKey, TValue> CreateKafkaConsumer(Action<IConsumer<TKey, TValue>, Error> consumeErrorHandler = null,
+                                                                           Action<IConsumer<TKey, TValue>, List<TopicPartition>> partitionAssignedHandler = null,
+                                                                           Action<IConsumer<TKey, TValue>, List<TopicPartitionOffset>> partitionRevokedHandler = null)
+        { 
+            return new Confluent.Kafka.ConsumerBuilder<TKey, TValue>(this.Properties)
```
Aaronontheweb: Looks good to me

src/Akka.Streams.Kafka/Settings/ProducerSettings.cs
```diff
@@ -71,10 +72,16 @@ public ProducerSettings(ISerializer<TKey> keySerializer, ISerializer<TValue> val
...
                 flushTimeout: config.GetTimeSpan("flush-timeout", TimeSpan.FromSeconds(2)),
-                properties: ImmutableDictionary<string, object>.Empty);
+                properties: ImmutableDictionary<string, string>.Empty);
         }
 
-        public Confluent.Kafka.IProducer<TKey, TValue> CreateKafkaProducer() =>
-            new Confluent.Kafka.Producer<TKey, TValue>(Properties, KeySerializer, ValueSerializer);
+        public Confluent.Kafka.IProducer<TKey, TValue> CreateKafkaProducer(Action<IProducer<TKey, TValue>, Error> producerErrorHandler = null)
+        {
+            return new Confluent.Kafka.ProducerBuilder<TKey, TValue>(Properties)
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -107,78 +101,29 @@ public override void PreStart()
...
-            _consumer.OnPartitionsAssigned -= HandleOnPartitionsAssigned;
-            _consumer.OnPartitionsRevoked -= HandleOnPartitionsRevoked;
-
             Log.Debug($"Consumer stopped: {_consumer.Name}");
             _consumer.Dispose();
 
             base.PostStop();
         }
+        
+        protected override void OnTimer(object timerKey) => PullQueue();
```
Aaronontheweb: So I know you're probably planning on changing this, but we don't actually need `OnTimer` or any sort of timer-driven behavior anymore now that we're just pulling from the Kafka topic on-demand, right?

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -192,23 +137,65 @@ private void MessagesReceived(ConsumerRecord<K, V> message)
...
+        
+        private void HandleConsumeError(object sender, Error error)
+        {
+            Log.Error(error.Reason);
+            var exception = new KafkaException(error);
+            switch (_decider(exception))
+            {
+                case Directive.Stop:
+                    // Throw
+                    _completion.TrySetException(exception);
```
Aaronontheweb: Looks good

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -192,23 +137,65 @@ private void MessagesReceived(ConsumerRecord<K, V> message)
...
+                    FailStage(exception);
+                    break;
+                case Directive.Resume:
+                    // keep going
+                    break;
+                case Directive.Restart:
+                    // keep going
```
Aaronontheweb: Might need to do something different here - for instance, if the issue is that we've lost connectivity to all Kafka hosts an we have to restart the client we'd want to fail this and get rebuilt, I suspect. Maybe the best place to handle that is the default supervision directive?

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -192,23 +137,65 @@ private void MessagesReceived(ConsumerRecord<K, V> message)
...
+                    return;
+                
+                HandleMessage(message);
+            }
+            catch (ConsumeException ex)
+            {
+                HandleError(ex.Error);
+            }
 
             if (!_isPaused && _buffer.Count > _settings.BufferSize)
```
Aaronontheweb: Guess we can get rid of the `_buffer` here?

src/Akka.Streams.Kafka/Messages/MessageAndMeta.cs
```diff
@@ -0,0 +1,16 @@
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Messages
+{
+    /// <summary>
+    /// Container for storing message and topic/parition info
+    /// </summary>
+    /// <typeparam name="K">Type of key</typeparam>
+    /// <typeparam name="V">Type of value</typeparam>
+    public class MessageAndMeta<K, V>
```
Aaronontheweb: Which parts of the solution are going to use this message type?

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -48,13 +48,13 @@ private async Task GivenInitializedTopic(string topic)
...
         {
-            return ConsumerSettings<Null, string>.Create(Sys, null, new StringDeserializer(Encoding.UTF8))
+            return ConsumerSettings<Null, string>.Create(Sys, null, null)
                 .WithBootstrapServers(KafkaUrl)
                 .WithProperty("auto.offset.reset", "earliest")
                 .WithGroupId(group);
         }
 
-        [Fact(Skip = "Needs IMPL")]
+        [Fact]
```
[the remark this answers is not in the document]
IgorFedchenko: Ok.
Actually, before making PR, I managed to make all tests pass. They require kafka to be up and running on local port 29092, so this is a first reason why they may fail on azure pipelines. Also, they still sometimes pass and sometimes fail locally - this is often the case when unit tests are not isolated and depend on execution order and shared state. And here the running container is shared state itself...
All in all, I will return to tests issue after rewriting them with Docker.NET.

src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
IgorFedchenko: Mostly because this CommitableOffset class is instantiated from consumer.Commit response like this:

``` csharp
var commitableOffset = new CommitableOffset(
    () => consumer.Commit(),
    new PartitionOffset("groupId", message.Topic, message.Partition, message.Offset));
```

And now Commit returns `List<TopicPartitionOffset>` instead of `CommittedOffsets`.

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: Right

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: I guess so - did not change this error handling logic, but indeed, if restart is required we will need to ask supervisor to restart. Will return to this later (maybe create some issue for this - how do you think?)

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: Sure, I will

src/Akka.Streams.Kafka/Messages/MessageAndMeta.cs
IgorFedchenko: Previously, this message type was build-in in Confluence.Kafka, and it was used to store message and destination (topic or topic partition info) before passing to kafka producer. So this is exclusively for Producer part.
Now new api require to pass topic or topic partition as a separate parameter to Produce method, along with message data. But it is just a matter of presentation, I think having a stream of `MessageAndMetadata`  objects is most simple way to pass messages and metadata.

So now I am doing the following to use required overload of Produce method:

``` csharp
onPush: () =>
{
    var msg = Grab(_stage.In);
    var result = new TaskCompletionSource<DeliveryReport<K, V>>();

    void PublishAction(Action<DeliveryReport<K, V>> report)
    {
        if (msg.TopicPartition != null)
            _producer.Produce(msg.TopicPartition, msg.Message, report);
        else
            _producer.Produce(msg.Topic, msg.Message, report);
    }

    PublishAction(report =>
    {
```

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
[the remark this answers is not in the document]
Aaronontheweb: >this is often the case when unit tests are not isolated and depend on execution order and shared state. And here the running container is shared state itself...

There are some XUnit features we can leverage to help with this.

https://github.com/akkadotnet/Akka.Persistence.SqlServer/blob/5f253418a33923f43f8ca32e996c7de06abd9de7/src/Akka.Persistence.SqlServer.Tests/SqlServerFixture.cs#L19-L22

If you use this in combination with the XUnit `IAsyncLifetime` object (running Kafka in Docker.NET), you can ensure that all tests from the `SqlServerSpec` collection all get access to the same container instance.

https://github.com/akkadotnet/Akka.Persistence.SqlServer/blob/5f253418a33923f43f8ca32e996c7de06abd9de7/src/Akka.Persistence.SqlServer.Tests/SqlServerJournalSpec.cs#L15-L16

When you decorate a class with the ` [Collection("SqlServerSpec")]` attribute, this means that all of the tests in that fixture will run sequentially, using the same shared `ICollectionFixture` object. The combination of those tools should help give you the ability to run those tests in a specific order. Does that help?

src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
Aaronontheweb: Ah got it - understood.

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
Aaronontheweb: Good, just wanted to understand

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
Aaronontheweb: > Will return to this later (maybe create some issue for this - how do you think?)

I think we can return to it later - this is something that can be overridden by users via Streams attributes and we should probably provide a default of some kind. Let's add a GH issue for it and get back to it.

src/Akka.Streams.Kafka/Messages/MessageAndMeta.cs
Aaronontheweb: Got it, that makes sense

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
realvictorprm: Hi, did you do anything in this area? We're experiencing the issue that the stream is in a zombie state because all brokers are not available yet. Any ideas?
@IgorFedchenko 
@Aaronontheweb

Aaronontheweb: Can you open an issue? @realvictorpm

Issue #39, opened 2019-08-14
Verify existing tests match alpakka-kafka project implementation and make our implementation pass tests


IgorFedchenko:
Currently, some tests are not stable - sometimes pass and sometimes fail. Also, I see some differences in logic between .net and scala implementations, which can lead to this unstable behavior. 

What we need here is:

1. Review tests and make their implementation match to original alpakka tests (at least on logical level). This does not include adding missing tests, but this also could be done.
2. Make our current implementation pass this tests in a stable way.

Issue #36, opened 2019-08-12
Implement missing stage types from alpakka-kafka project


IgorFedchenko:
There are lots of consumer and provider stage types already implemented in this project: https://github.com/akka/alpakka-kafka

We need to implement consumers from here:
https://doc.akka.io/docs/alpakka-kafka/current/consumer.html
And producers from here:
https://doc.akka.io/docs/alpakka-kafka/current/producer.html

Some of them are going to be based on brand new `flowWithContext`-type stages implemented, so until `Akka.NET v1.4.0-beta2` will be released, this can be based on nightly Akka builds.

This is not critical to implement all stages from alpakka right now, so the concrete must-have list to implement will be specified/discussed later.

IgorFedchenko:
I wrote some short overview about how current scala implementation uses buffering and separate actor for stages implementation [here](https://github.com/akkadotnet/Akka.Streams.Kafka/issues/35#issuecomment-521247203)

It looks like we will need to implement something similar in our project. This is going to be a series of PRs with 
1. Code refactoring to use base classes containing shared logic between stages.
2. Implementing of missing stages (or some of them)
3. Looking at the external actor usage to consume messages in batch.

The one more thing I noticed in scala [imlementation](https://github.com/akka/alpakka-kafka/blob/master/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala) is that `consumer.Poll` method they use returns list of messages - but our confluent driver returns only one message in a time. This may result in somewhat different implementation on our side.

Aaronontheweb:
@IgorFedchenko let's add the following stages next so we can do a beta release of this package:

*  FlowWithContext
* Flow
* Sink

IgorFedchenko:
@Aaronontheweb So you want to implement some of Transactional Producers?
Also, I suspect I will need `FlexiFlow` producer implemented to implement `FlowWithContext` - I will check that out.

Aaronontheweb:
Ok, got it - thank you!

Aaronontheweb:
@IgorFedchenko so only 3 more stages left?

IgorFedchenko:
@Aaronontheweb Technically yes. But the thing is that transactional sources are ment to be used in pair with transactional producers. And transactional producers are based on `IProducer` of Confluent driver, which does not support transactions yet (see our issue #85 ). So, until they will add support for transactions, there is no much point in implementing the rest of the stages.

[...]

PR #40, opened 2019-08-14
Remove timer and buffering from source stages

IgorFedchenko:
This PR is assigned to issue #35 . 

Most changes are duplicating between ConsumerStage and CommittableConsumerStage - that is because they are very similar. We should refactor them to inherit common logic from some base class  - well, how this is implemented in alpakka-kafka project (so this is going to be a part of issue #36 I guess)

At least one of the tests is still failing - this one:
https://github.com/akkadotnet/Akka.Streams.Kafka/blob/c609061bab16e846f953dcc605b90627a58d3544/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs#L57-L58

I checked out it's implementation in alpakka and there is a difference which may lead to race conditions and fail tests. This will be solved in issue #39 .
Also, in this PR all tests should fail because this update is independent from PR #38 and thus does not use Docker.NET to run containers during tests.

### DIFF
--- a/src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
+++ b/src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
@@ -8,6 +8,7 @@ using Akka.Streams.Stage;
 using Confluent.Kafka;
 using Akka.Streams.Supervision;
 using System.Runtime.Serialization;
+using System.Threading;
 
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@IgorFedchenko can we close this PR since it was part of #41?

Aaronontheweb:
Resolved via #41

PR #41, opened 2019-08-15
Make tests pass

### LINKED ISSUE #39: Verify existing tests match alpakka-kafka project implementation and make our implementation pass tests

IgorFedchenko:
This PR actually contains two currently pending pull requests: #38 and #40 .

#38 moved tests to Docker.NET and basically makes CI process working
#40 Updates implementation to not use buffering and timers in consumer stages. But CI will not show this PR passing, because it is missing Docker.NET 

Also, even with #38 and #40 applied, one of tests was still failing sometimes. So, the goal of this PR is to make current implementation pass tests. This will allow us to safely move further with working (hence incomplete) test coverage.

To review this, please ignore commits before merges - they are from other PRs. The actual code to review here is pretty small, it fixes test to match alpakka implementation (as fixes race conditions as well), and sets KafkaConsumer source task as completed after stage was stopped.

This also can possibly close issue #39 , because I was creating it with this PR in mind:
Close #39 

P.S. Let me know if I should not create PRs based on other PRs until they are merged.

P.P.S. While tests are passing, sometimes locally I have a weird issue that tests are freezing in some point, and executing forever. This is not easy to reproduce, so will return to this if  the problem will remain until the end of milestone.

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -12,18 +12,18 @@ using Akka.Streams.TestKit;
 using Confluent.Kafka;
 using Xunit;
 using Xunit.Abstractions;
+using Config = Akka.Configuration.Config;
 
 namespace Akka.Streams.Kafka.Tests.Integration
 {
-    [Collection(KafkaSpecsFixture.Name)]
-    public class CommittableSourceIntegrationTests : Akka.TestKit.Xunit2.TestKit
+    public class CommittableSourceIntegrationTests : KafkaIntegrationTests
     {
         private const string InitialMsg = "initial msg in topic, required to create the topic before any consumer subscribes to it";
         private readonly KafkaFixture _fixture;
         private readonly ActorMaterializer _materializer;
 
         public CommittableSourceIntegrationTests(ITestOutputHelper output, KafkaFixture fixture) 
-            : base(ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf"), null, output)
+            : base(nameof(CommittableSourceIntegrationTests), output)
         {
             _fixture = fixture;
             _materializer = Sys.Materializer();
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -101,7 +101,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
             var consumerSettings = CreateConsumerSettings(group1);
             var committedElements = new ConcurrentQueue<string>();
 
-            var (_, probe1) = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
+            var (task, probe1) = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
                 .WhereNot(c => c.Record.Value == InitialMsg)
                 .SelectAsync(10, elem =>
                 {
@@ -121,7 +121,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
                 
             probe1.Cancel();
 
-            // Await.result(control.isShutdown, remainingOrDefault)
+            AwaitCondition(() => task.IsCompleted);
 
             var probe2 = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
                 .Select(_ => _.Record.Value)
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
@@ -15,17 +15,17 @@ using Xunit.Abstractions;
 
 namespace Akka.Streams.Kafka.Tests.Integration
 {
+    [Collection(KafkaSpecsFixture.Name)]
     public class CommittableSourceIntegrationTests : Akka.TestKit.Xunit2.TestKit
     {
-        private const string KafkaUrl = "localhost:29092";
-
         private const string InitialMsg = "initial msg in topic, required to create the topic before any consumer subscribes to it";
-
+        private readonly KafkaFixture _fixture;
         private readonly ActorMaterializer _materializer;
 
-        public CommittableSourceIntegrationTests(ITestOutputHelper output) 
+        public CommittableSourceIntegrationTests(ITestOutputHelper output, KafkaFixture fixture) 
             : base(ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf"), null, output)
         {
+            _fixture = fixture;
             _materializer = Sys.Materializer();
         }
 
@@ -34,9 +34,10 @@ namespace Akka.Streams.Kafka.Tests.Integration
         private string CreateTopic(int number) => $"topic-{number}-{Uuid}";
         private string CreateGroup(int number) => $"group-{number}-{Uuid}";
 
-        private ProducerSettings<Null, string> ProducerSettings =>
-            ProducerSettings<Null, string>.Create(Sys, null, null)
-                .WithBootstrapServers(KafkaUrl);
+        private ProducerSettings<Null, string> ProducerSettings
+        {
+            get => ProducerSettings<Null, string>.Create(Sys, null, null).WithBootstrapServers(_fixture.KafkaServer);
+        }
 
         private async Task GivenInitializedTopic(string topic)
         {
@@ -49,7 +50,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
         private ConsumerSettings<Null, string> CreateConsumerSettings(string group)
         {
             return ConsumerSettings<Null, string>.Create(Sys, null, null)
-                .WithBootstrapServers(KafkaUrl)
+                .WithBootstrapServers(_fixture.KafkaServer)
                 .WithProperty("auto.offset.reset", "earliest")
                 .WithGroupId(group);
         }
--- a/src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
+++ b/src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
@@ -1,0 +1,165 @@
+using System;
+using System.Collections.Generic;
+using System.Data.Common;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using Akka.Util;
+using Docker.DotNet;
+using Docker.DotNet.Models;
+using Xunit;
+
+namespace Akka.Streams.Kafka.Tests
+{
+    [CollectionDefinition(Name)]
+    public sealed class KafkaSpecsFixture : ICollectionFixture<KafkaFixture>
+    {
+        public const string Name = "KafkaSpecs";
+    }
+    
+    public class KafkaFixture : IAsyncLifetime
+    {
+        private const string KafkaImageName = "confluentinc/cp-kafka";
+        private const string KafkaImageTag = "latest";
+        private const string ZookeeperImageName = "confluentinc/cp-zookeeper";
+        private const string ZookeeperImageTag = "latest";
+        
+        private readonly string _kafkaContainerName = $"kafka-{Guid.NewGuid():N}";
+        private readonly string _zookeeperContainerName = $"zookeeper-{Guid.NewGuid():N}";
+        private readonly string _networkName = $"network-{Guid.NewGuid():N}";
+        
+        private readonly DockerClient _client;
+        
+        public KafkaFixture()
+        {
+            DockerClientConfiguration config;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                config = new DockerClientConfiguration(new Uri("unix://var/run/docker.sock"));
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                config = new DockerClientConfiguration(new Uri("npipe://./pipe/docker_engine"));
+            else
+                throw new NotSupportedException($"Unsupported OS [{RuntimeInformation.OSDescription}]");
+
+            _client = config.CreateClient();
+        }
+        
+        public int KafkaPort { get; private set; }
+        public string KafkaServer => $"localhost:{KafkaPort}";
+
+        public async Task InitializeAsync()
+        {
+            // Load images, if they not exist yet
+            await EnsureImageExists(ZookeeperImageName, ZookeeperImageTag);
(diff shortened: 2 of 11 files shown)


IgorFedchenko:
CI tests failed due to process timed out... It looks like this is the problem with tests freezing I have mentioned above. All right, we need this tests working in this PR, so will continue committing fixes until tests will pass.

IgorFedchenko:
So, in last 4 commits I did the following:
1) Because I did not saw output in console until test finished, and the issue was that test did not finish at all, I implemented simple akka logger to log all lines per test in separate file. Called it `SimpleFileLogger`.

2) To enable this logger I had to set it in each tests class configuration - so moved akka configuration stuff to base class - also this allows to set xunit `[Collection]` attribute to all child classes out of the box.

3) The actual test fix: turned out that is was ProducerStage that was a reason of hanging tests. There is a `CheckIfCompleted` method that was called concurrently from two threads - the own stage thread and from kafka procude action callback. And - this method was calling `CompleteStage` method of `GraphStageLogic `class and it was not doing anything when called from another thread. So, I made this call threadsafe with `GetAsyncCallback` built-in method, and that's it.

4) While I was going to just disable this file logging with "false" constant, this is more convenient to use some environment setting, so I set it to check `AKKA_STREAMS_KAFKA_FILE_LOGGING` environment variable. If you will set it for testing on your local machine, you will get all logs duplicated to hard disk "logs" folder near tests assembly.

The goal of this PR is achieved now. All tests stably pass.

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -12,20 +12,20 @@
...
-        private const string KafkaUrl = "localhost:29092";
-
         private const string InitialMsg = "initial msg in topic, required to create the topic before any consumer subscribes to it";
-
+        private readonly KafkaFixture _fixture;
         private readonly ActorMaterializer _materializer;
 
-        public CommittableSourceIntegrationTests(ITestOutputHelper output) 
-            : base(ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf"), null, output)
+        public CommittableSourceIntegrationTests(ITestOutputHelper output, KafkaFixture fixture)
```
Aaronontheweb: Looks good to me!

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -120,7 +121,7 @@ await Source
                 
             probe1.Cancel();
 
-            // Await.result(control.isShutdown, remainingOrDefault)
+            AwaitCondition(() => task.IsCompleted);
```
Aaronontheweb: :+1:

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+    [CollectionDefinition(Name)]
+    public sealed class KafkaSpecsFixture : ICollectionFixture<KafkaFixture>
+    {
+        public const string Name = "KafkaSpecs";
+    }
+    
+    public class KafkaFixture : IAsyncLifetime
+    {
+        private const string KafkaImageName = "confluentinc/cp-kafka";
+        private const string KafkaImageTag = "latest";
```
Aaronontheweb: @IgorFedchenko let's do this in a separate PR, but I would lock down the Docker image tag version to something specific - I'm looking at doing this with some of our other projects too. It's not an issue for most updates to the Docker image versions, but when there's a major release that includes some breaking changes that can cause CI to fail by accident. We should require future users to have to do that intentionally, usually at the same time as driver upgrades. Does that make sense?

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+        private const string ZookeeperImageName = "confluentinc/cp-zookeeper";
+        private const string ZookeeperImageTag = "latest";
```
Aaronontheweb: Same feedback as for the Kafka tag

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+            var zookeeperPort = ThreadLocalRandom.Current.Next(32000, 33000);
+            KafkaPort = ThreadLocalRandom.Current.Next(28000, 29000);
+
+            // create the containers
+            await CreateContainer(ZookeeperImageName, ZookeeperImageTag, _zookeeperContainerName, zookeeperPort, new Dictionary<string, string>()
+            {
+                ["ZOOKEEPER_CLIENT_PORT"] = zookeeperPort.ToString(),
+                ["ZOOKEEPER_TICK_TIME"] = "2000",
+            });
+            await CreateContainer(KafkaImageName, KafkaImageTag, _kafkaContainerName, KafkaPort, new Dictionary<string, string>()
```
Aaronontheweb: This looks right

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+                            new List<PortBinding>
+                            {
+                                new PortBinding
+                                {
+                                    HostPort = $"{portToExpose}"
+                                }
+                            }
+                        }
+                    },
+                    ExtraHosts = new [] { "moby:127.0.0.1", "localhost:127.0.0.1" },
```
Aaronontheweb: What's the `moby:127.0.0.1` for? Is this for Docker on Windows?

src/Akka.Streams.Kafka.Tests/KafkaIntegrationTests.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using Akka.Configuration;
+using Akka.Streams.Kafka.Settings;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Kafka.Tests
+{
+    [Collection(KafkaSpecsFixture.Name)]
+    public class KafkaIntegrationTests : Akka.TestKit.Xunit2.TestKit
```
Aaronontheweb: Good idea

src/Akka.Streams.Kafka.Tests/Logging/SimpleFileLoggerActor.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using System.IO;
+using Akka.Actor;
+using Akka.Event;
+
+namespace Akka.Streams.Kafka.Tests.Logging
+{
+    public class SimpleFileLoggerActor : ReceiveActor
```
Aaronontheweb: Is this just for debugging purposes?

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -32,51 +33,56 @@ public override ILogicAndMaterializedValue<Task> CreateLogicAndMaterializedValue
...
-                if (_buffer.Count > 0)
-                {
-                    Push(_out, _buffer.Dequeue());
-                }
-                else
+                try
                 {
-                    if (_isPaused)
+                    var message = _consumer.Consume(_cancellationTokenSource.Token);
+                    if (message == null) // No message received, or consume error occured
```
Aaronontheweb: Good

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -32,51 +33,56 @@ public override ILogicAndMaterializedValue<Task> CreateLogicAndMaterializedValue
...
+                        return;
+
+                    var consumer = _consumer;
+                    var commitableOffset = new CommitableOffset(
```
Aaronontheweb: Does the `groupId` string need to be hard coded?

Aaronontheweb: And if so, why? Asking because I don't know enough about the Kafka driver to fully understand myself.

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -32,51 +33,56 @@ public override ILogicAndMaterializedValue<Task> CreateLogicAndMaterializedValue
...
-                        _isPaused = false;
-                        Log.Debug("Polling resumed, buffer is empty");
+                        var commitableMessage = new CommittableMessage<K, V>(message, commitableOffset);
+                        Push(stage.Out, commitableMessage);
                     }
-                    PullQueue();
+                }
+                catch (OperationCanceledException)
+                {
+                    // Consume was canceled, looks like we are shutting down the stage
```
Aaronontheweb: Is there something we need to do to cancel the stage here? Or is there already going to be a stage termination in progress by the time this exception is thrown? And is `OperationCanceledException` thrown when the `CancellationToken` is invoked or is that going to throw a `TaskCancellationException` ?

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -101,58 +107,39 @@ public override void PreStart()
...
             _partitionsAssigned = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsAssigned);
             _partitionsRevoked = GetAsyncCallback<IEnumerable<TopicPartitionOffset>>(PartitionsRevoked);
-            ScheduleRepeatedly(TimerKey, _settings.PollInterval);
         }
 
         public override void PostStop()
         {
             Log.Debug($"Consumer stopped: {_consumer.Name}");
             _consumer.Dispose();
+            _completion.SetResult(NotUsed.Instance);
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
```diff
@@ -101,58 +107,39 @@ public override void PreStart()
...
-        protected override void OnTimer(object timerKey) => PullQueue();
-
         //
         // Consumer's events
         //
 
-        private void HandleMessage(ConsumeResult<K, V> message) => _messagesReceived(message);
-        
-        private void MessagesReceived(ConsumeResult<K, V> message)
+        private void HandleOnPartitionsAssigned(IConsumer<K, V> consumer, List<TopicPartition> list)
```
Aaronontheweb: When does this method get assigned?

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
IgorFedchenko: Sure, you are right. Actually, I was going to have it fixed for the same reason, but saw that "latest" tag is used for SQLServer fixture in reference project, so made it the same way.

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: Indeed, did not notice this, left as it was... Thanks for noticing! GroupId is the thing we may pass to consumer factory and is constant for all messages we consume - it is stored in _settings stage field, so I will pass real value instead of hard-coded value in separate PR.

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
IgorFedchenko: Mm, this is something I just took from docker-compose.yml file that was used before to make setup in the same way it was working before. I do not see any reference about it in official image documentation, so will try to remove it and check tests passing in separate PR.

src/Akka.Streams.Kafka.Tests/Logging/SimpleFileLoggerActor.cs
IgorFedchenko: Yes. There is a `UseFileLogging` option in `KafkaIntegrationTests` class:

```csharp
/// <summary>
/// Allows to write logs to file (useful for debugging when tests are running forever and no output in console is available)
/// </summary>
private static readonly bool UseFileLogging = Environment.GetEnvironmentVariable("AKKA_STREAMS_KAFKA_FILE_LOGGING") != null;
```

So this logger is used if this option is set:
```csharp
if (UseFileLogging)
{
	config = config.WithFallback(ConfigurationFactory.ParseString("akka.loggers = [\"Akka.Streams.Kafka.Tests.Logging.SimpleFileLoggerActor, Akka.Streams.Kafka.Tests\"]"));
}
```

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: I am using it to handle `CancellationToken` usage, it was throwing `OperationCanceledException` when cancellation was requested. This cancellation is possible only when we get consume error and `HandleConsumeError` handler is invoked (currently on separate thread my the way). But may be this is because usually `HandleConsumeError` is called before next onPull is called and at the moment we call `_consumer.Consume(_cancellationTokenSource.Token)` token is already canceled and we get `OperationCanceledException`. And if for some reason  `_consumer.Consume(_cancellationTokenSource.Token)` will be called before cancellation is requested, we will get `TaskCancellationException`, because the consuming task will be already started internally... I will check this out and make update in separate PR. Possibly catching `TaskCancellationException` as well will be enough.

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
IgorFedchenko: Sorry, did not get it: which method do you mean? `HandleMessage` method is not needed anymore, because we are not buffering right now and just pushing message to the downstream.
And `HandleOnPartitionsAssigned` method is used as a callback for `partitionsAssigned` consumer event, it is calling "_partitionsAssigned" thread-safe delegate and that delegate is a wrapper for `PartitionsAssigned` method. Like this:

```csharp
public override void PreStart()
{
	// ...
	_consumer = _settings.CreateKafkaConsumer(..., HandlePartitionsAssigned, ...);
	// ...
	_partitionsAssigned = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsAssigned);
	// ...
}


private void HandlePartitionsAssigned(IConsumer<K, V> consumer, List<TopicPartition> list)
{
	_partitionsAssigned(list);
}

private void PartitionsAssigned(IEnumerable<TopicPartition> partitions)
{
	// Real handling is implemented here
}
```

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
Aaronontheweb: We'll see how things go during integration testing

src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs
Aaronontheweb: Ahhh I see.

PR #45, opened 2019-08-24
Refactor to make consumer stages to use base classes

IgorFedchenko:
This work is a part of issue #36 .

Alpakka project uses base classes to implement common logic between stages. There are lots of abstractions, and even more would be required to be implemented in c# because scala has mixin feature and allows developers to do kind of inline inheritance (like specifying base class and overrides in same place as a part of inheritance syntax) - and in c# we need to implement separate child class each time and use it instead.

This is a first PR in the series of PRs in issue #36 , and the goal of this one is to:

1. Use `KafkaSourceStage` base class for all consumer stages definition
2. Use `SingleSourceStageLogic` base class of some of consumer stage logic classes. So far we have `PlainSourceStage` and `CommittableSourceStage`, and both of them inherit from this class.

While common logic of message consuming and handling is moved to `SingleSourceStageLogic` class, the difference is in downstream messages - `PlainSourceStage` pushes plain consumed records, while `CommittableSourceStage` pushes `CommittableMessage`'s, which allows committing offsets when desired. To support this difference, there is a `IMessageBuilder` interface and different implementations, so far only `PlainMessageBuilder` and `CommittableSourceMessageBuilder` are implemented.

To simplify the review, I will make review and put some comments on changes right now.


src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -72,6 +72,15 @@ await CreateContainer(KafkaImageName, KafkaImageTag, _kafkaContainerName, KafkaP
...
             });
 
+            /*
+            // Remove old networks, if there are any
+            var networks = await _client.Networks.ListNetworksAsync(new NetworksListParameters());
+            foreach (var existingNetwork in networks.Where(n => n.Name.StartsWith("network-")))
+            {
+                await _client.Networks.DeleteNetworkAsync(existingNetwork.ID);
+            }
+            */
```
IgorFedchenko: This part is mostly for local debugging.
When tests are aborted, no cleanup code is executed and some resources are still sitting in the system.
While removing containers is pretty simple using docker Kitemaric UI, networks cleanup should be performed manually. So, this code part helps.
Maybe in future I will create PR performing cleanup on tests setup before allocating new resources.

src/Akka.Streams.Kafka/Dsl/KafkaConsumer.cs
```diff
@@ -24,7 +25,7 @@ public static class KafkaConsumer
         /// </summary>
         public static Source<ConsumeResult<K, V>, Task> PlainSource<K, V>(ConsumerSettings<K, V> settings, ISubscription subscription)
         {
-            return Source.FromGraph(new KafkaSourceStage<K, V>(settings, subscription));
+            return Source.FromGraph(new PlainSourceStage<K, V>(settings, subscription));
```
IgorFedchenko: Some renaming to make it more similar to alpakka naming and thus making it simpler to compare/verify implementation.

src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
```diff
@@ -30,21 +41,33 @@ public CommittableMessage(ConsumeResult<K, V> record, CommitableOffset commitabl
...
+            Metadata = metadata;
         }
 
-        public PartitionOffset Offset { get; }
-
-        public List<TopicPartitionOffset> Commit()
+        public Task Commit()
         {
-            return _task();
+            return Task.FromResult(_committer.Commit());
```
IgorFedchenko: Alpakka's Commit method returns `Future[Done]`, and committed partition offsets can be accessed from `CommittableMessage` directly, so no need to return them from Commit method. Also, in different stages there are different implementations of `IInternalCommitter` `commit `method, so more general `Task` response is better here.

Also, the `Func<List<TopicPartitionOffset>> task` field is replaced with `IInternalCommitter committer` field to allow using additional info for commitment as alpakka does (and we will use this flexibility in `KafkaInternalCommitter` class).

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -48,27 +25,30 @@ internal class KafkaSourceStageLogic<K, V> : GraphStageLogic
         private readonly TaskCompletionSource<NotUsed> _completion;
         private readonly CancellationTokenSource _cancellationTokenSource;
 
-        public KafkaSourceStageLogic(KafkaSourceStage<K, V> stage, Attributes attributes, TaskCompletionSource<NotUsed> completion) : base(stage.Shape)
+        public SingleSourceStageLogic(SourceShape<TMessage> shape, ConsumerSettings<K, V> settings,
```
IgorFedchenko: This source stage was transformed to more general `SingleSourceStage`, that uses `IMessageBuilder` argument to pass consumed messages to downstream and does not depend on concrete stage like (because only shape of the stage is required to implement logic, not the stage itself).

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -48,27 +25,30 @@ internal class KafkaSourceStageLogic<K, V> : GraphStageLogic
...
                 {
                     var message = _consumer.Consume(_cancellationTokenSource.Token);
                     if (message == null) // No message received, or consume error occured
                         return;
 
-                    if (IsAvailable(stage.Out))
+                    if (IsAvailable(shape.Outlet))
                     {
-                        Push(stage.Out, message);
+                        Push(shape.Outlet, messageBuilder.CreateMessage(message, _consumer));
```
IgorFedchenko: Using generic `IMessageBuilder` to build downstream message.
The tricky part here is passing `_consumer` instance to builder. Alpakka does not do that. Instead, they use GraphStage's `materializer` property to get consumer actor reference and use it in message builder like this:

```scala
  override protected def logic(shape: SourceShape[CommittableMessage[K, V]]): GraphStageLogic with Control =
    new SingleSourceLogic[K, V, CommittableMessage[K, V]](shape, settings, subscription)
    with CommittableMessageBuilder[K, V] {
      override def metadataFromRecord(record: ConsumerRecord[K, V]): String = _metadataFromRecord(record)
      override def groupId: String = settings.properties(ConsumerConfig.GROUP_ID_CONFIG)
      lazy val committer: KafkaAsyncConsumerCommitterRef = {
        val ec = materializer.executionContext
        new KafkaAsyncConsumerCommitterRef(consumerActor, settings.commitTimeout)(ec)
      }
    }
```

We do not use any consumer actor right now, neither Akka.Streams gives `materializer` property - or maybe I did not find it - so to allow builder to execute commit action, I am passing consumer instance directly.

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
IgorFedchenko: Added issue #46 to keep track on this.

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Generic;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages.Consumers
+{
+    public interface IInternalCommitter
```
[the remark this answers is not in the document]
IgorFedchenko: No, you are right, this is internal API - will fix this

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
```diff
@@ -0,0 +1,28 @@
...
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages.Consumers
+{
+    public interface IInternalCommitter 
+    {
+        /// <summary>
+        /// Commit all offsets (of different topics) belonging to the same stage
+        /// </summary>
+        List<TopicPartitionOffset> Commit();
```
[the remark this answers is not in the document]
IgorFedchenko: I do not think so - it is kafka's driver returning mutable collection, but indeed, there is no point in such mutations. Will update this

src/Akka.Streams.Kafka/Stages/Consumers/Concrete/CommittableSourceStage.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Threading.Tasks;
+using Akka.Streams.Kafka.Messages;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.Stage;
+using Akka.Streams.Supervision;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages.Consumers
+{
+    internal class CommittableSourceStage<K, V> : KafkaSourceStage<K, V, CommittableMessage<K, V>>
```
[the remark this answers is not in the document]
IgorFedchenko: Done

src/Akka.Streams.Kafka/Stages/Consumers/Concrete/PlainSourceStage.cs
```diff
@@ -0,0 +1,26 @@
+using System.Threading.Tasks;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.Stage;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages.Consumers
+{
+    internal class PlainSourceStage<K, V> : KafkaSourceStage<K, V, ConsumeResult<K, V>>
```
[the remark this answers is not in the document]
IgorFedchenko: Done

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    /// <summary>
+    /// Interface for implementing committing consumed messages
+    /// </summary>
+    internal interface IInternalCommitter 
+    {
+        /// <summary>
+        /// Commit all offsets (of different topics) belonging to the same stage
+        /// </summary>
+        Task Commit();
```
IgorFedchenko: @Aaronontheweb Alpakka returnes `Task `here instead of topic partitions list as I did before - and this makes sense, because the public `ICommittable` interface returns general `Task` without typed value.

IgorFedchenko:
Sorry, I see some conflicts with dev branch - will resolve now

IgorFedchenko:
Ok, now ready for review.

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
Aaronontheweb: I think that's fine

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
Aaronontheweb: @IgorFedchenko the materializer might be passed in as a Scala `implicit` - which makes the code somewhat hard to read. I'll take a look and see how they're doing this.

Aaronontheweb: Or maybe you might have some other ideas

PR #47, opened 2019-09-04
Added resource cleanup on start/stop testing

IgorFedchenko:
This PR is about tests cleanup and should make life easier during local development.  It addresses issue #46 .

Close #46

All docker calls during cleanup are executed in parallel when possible, to decrease start/stop delays.

PR #48, opened 2019-09-06
Base stage with consuming actor

IgorFedchenko:
This PR is related to issue #36 . 

To move on and add different types of stages for consumers, we need to add two more abstractions:
 
1. `BaseSingleSourceLogic` - this is a new subclass for `SingleSourceStageLogic`. While `SingleSourceStageLogic` still contains some configuration and callback handlers, new `BaseSingleSourceLogic` is used to handle messages consuming from consumer actor
2. `KafkaConsumerActor` - is an actor, that is used to manage kafka consumer client. Basically, it receives all subscriptions from stages, subscribes to all topics they are interested in, and consumes messages on demand, broadcasting them to requesters once consumed. Error handling is also performed in this actor, to stages now only responsible for handling error messages and handling them according to supervision strategy.

I am creating this PR as a job-in-progress draft - right know new implementation is failing all consuming tests. So I will fix it and then move to review.

Also, code changes are pretty big, so will try to put my notes here as a part of PR review to simplify code review.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -6,5 +6,6 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="Akka.NET Nightly" value="https://www.myget.org/F/akkadotnet/api/v2" />
   </packageSources>
 </configuration>
--- a/src/Akka.Streams.Kafka/Akka.Streams.Kafka.csproj
+++ b/src/Akka.Streams.Kafka/Akka.Streams.Kafka.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <AssemblyTitle>Akka.Streams.Kafka</AssemblyTitle>
     <Description>Apache Kafka adapter for Akka.NET Streams</Description>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <VersionPrefix>0.5.0</VersionPrefix>
     <VersionSuffix>beta</VersionSuffix>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Akka.Streams" Version="$(AkkaVersion)" />
+    <PackageReference Include="Akka.Streams" Version="1.4.0-beta637012445454141588" />
     <PackageReference Include="Confluent.Kafka" Version="1.1.0" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
   </ItemGroup>
--- a/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
+++ b/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
@@ -1,0 +1,34 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace Akka.Streams.Kafka.Extensions
+{
+    public static class CollectionExtensions
+    {
+        /// <summary>
+        /// Joins elements of the collection with given separator to single string
+        /// </summary>
+        public static string JoinToString<T>(this IEnumerable<T> collection, string separator)
+        {
+            return string.Join(separator, collection);
+        }
+
+        /// <summary>
+        /// Converts dictionary to a list of tuples
+        /// </summary>
+        public static IEnumerable<(TKey, TValue)> ToTuples<TKey, TValue>(this IEnumerable<KeyValuePair<TKey, TValue>> dictionary)
+        {
+            return dictionary.Select(pair => (pair.Key, pair.Value));
+        }
+
+        /// <summary>
+        /// Checks if collection is empty
+        /// </summary>
+        public static bool IsEmpty<T>(this IEnumerable<T> collection)
+        {
+            return !collection.Any();
+        }
+    }
+}
--- a/src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
+++ b/src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Threading.Tasks;
 using Akka.Streams.Kafka.Dsl;
 using Akka.Streams.Kafka.Stages.Consumers;
@@ -98,7 +99,7 @@ namespace Akka.Streams.Kafka.Messages
         /// </summary>
         public Task Commit()
         {
-            return _committer.Commit();
+            return _committer.Commit(new List<PartitionOffset>() { Offset }.ToImmutableList());
         }
     }
 
--- a/src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
+++ b/src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
@@ -24,6 +24,8 @@ namespace Akka.Streams.Kafka.Settings
                 valueDeserializer: valueDeserializer,
                 pollInterval: config.GetTimeSpan("poll-interval", TimeSpan.FromMilliseconds(50)),
                 pollTimeout: config.GetTimeSpan("poll-timeout", TimeSpan.FromMilliseconds(50)),
+                commitTimeout: config.GetTimeSpan("commit-timeout", TimeSpan.FromMilliseconds(50)),
+                stopTimeout: config.GetTimeSpan("stop-timeout", TimeSpan.FromMilliseconds(50)),
                 bufferSize: config.GetInt("buffer-size", 50),
                 dispatcherId: config.GetString("use-dispatcher", "akka.kafka.default-dispatcher"),
                 properties: ImmutableDictionary<string, string>.Empty);
@@ -35,16 +37,22 @@ namespace Akka.Streams.Kafka.Settings
         public IDeserializer<TValue> ValueDeserializer { get; }
         public TimeSpan PollInterval { get; }
         public TimeSpan PollTimeout { get; }
+        public TimeSpan CommitTimeout { get; }
+        public TimeSpan StopTimeout { get; }
         public int BufferSize { get; }
         public string DispatcherId { get; }
         public IImmutableDictionary<string, string> Properties { get; }
 
-        public ConsumerSettings(IDeserializer<TKey> keyDeserializer, IDeserializer<TValue> valueDeserializer, TimeSpan pollInterval, TimeSpan pollTimeout, int bufferSize, string dispatcherId, IImmutableDictionary<string, string> properties)
+        public ConsumerSettings(IDeserializer<TKey> keyDeserializer, IDeserializer<TValue> valueDeserializer, TimeSpan pollInterval, 
+                                TimeSpan pollTimeout, TimeSpan commitTimeout, TimeSpan stopTimeout, int bufferSize, 
+                                string dispatcherId, IImmutableDictionary<string, string> properties)
         {
             KeyDeserializer = keyDeserializer;
             ValueDeserializer = valueDeserializer;
             PollInterval = pollInterval;
             PollTimeout = pollTimeout;
+            StopTimeout = stopTimeout;
+            CommitTimeout = commitTimeout;
             BufferSize = bufferSize;
             DispatcherId = dispatcherId;
             Properties = properties;
@@ -65,6 +73,9 @@ namespace Akka.Streams.Kafka.Settings
         public ConsumerSettings<TKey, TValue> WithPollInterval(TimeSpan pollInterval) => Copy(pollInterval: pollInterval);
 
         public ConsumerSettings<TKey, TValue> WithPollTimeout(TimeSpan pollTimeout) => Copy(pollTimeout: pollTimeout);
+        
+        public ConsumerSettings<TKey, TValue> WithCommitTimeout(TimeSpan commitTimeout) => Copy(commitTimeout: commitTimeout);
+        public ConsumerSettings<TKey, TValue> WithStopTimeout(TimeSpan stopTimeout) => Copy(stopTimeout: stopTimeout);
 
         public ConsumerSettings<TKey, TValue> WithDispatcher(string dispatcherId) => Copy(dispatcherId: dispatcherId);
         
@@ -75,6 +86,8 @@ namespace Akka.Streams.Kafka.Settings
             IDeserializer<TValue> valueDeserializer = null,
             TimeSpan? pollInterval = null,
             TimeSpan? pollTimeout = null,
+            TimeSpan? commitTimeout = null,
+            TimeSpan? stopTimeout = null,
             int? bufferSize = null,
             string dispatcherId = null,
             IImmutableDictionary<string, string> properties = null) =>
@@ -83,6 +96,8 @@ namespace Akka.Streams.Kafka.Settings
                 valueDeserializer: valueDeserializer ?? this.ValueDeserializer,
                 pollInterval: pollInterval ?? this.PollInterval,
                 pollTimeout: pollTimeout ?? this.PollTimeout,
+                commitTimeout: commitTimeout ?? this.CommitTimeout,
+                stopTimeout: stopTimeout ?? this.StopTimeout,
                 bufferSize: bufferSize ?? this.BufferSize,
                 dispatcherId: dispatcherId ?? this.DispatcherId,
                 properties: properties ?? this.Properties);
--- a/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
+++ b/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
@@ -1,0 +1,159 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Dispatch;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.Kafka.Stages.Consumers.Actors;
+using Akka.Streams.Kafka.Stages.Consumers.Exceptions;
+using Akka.Streams.Stage;
+using Akka.Util;
+using Akka.Util.Internal;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
+{
+    /// <summary>
+    /// Shared GraphStageLogic for <see cref="SingleSourceStageLogic{K,V,TMessage}"/> and <see cref="ExternalSingleSourceLogic"/>
+    /// </summary>
+    /// <typeparam name="K">Key type</typeparam>
+    /// <typeparam name="V">Value type</typeparam>
+    /// <typeparam name="TMessage">Message type</typeparam>
+    internal abstract class BaseSingleSourceLogic<K, V, TMessage> : GraphStageLogic
+    {
+        private readonly SourceShape<TMessage> _shape;
+        private readonly TaskCompletionSource<NotUsed> _completion;
+        private readonly IMessageBuilder<K, V, TMessage> _messageBuilder;
+        private int _requestId = 0;
+        private bool _requested = false;
+        private readonly ConcurrentQueue<ConsumeResult<K, V>> _buffer = new ConcurrentQueue<ConsumeResult<K, V>>();
+        protected IImmutableSet<TopicPartition> TopicPartitions { get; set; } = ImmutableHashSet.Create<TopicPartition>();
+        
+        protected StageActor SourceActor { get; private set; }
+        internal IActorRef ConsumerActor { get; private set; }
+        internal MessageDispatcher ExecutionContext => Materializer.ExecutionContext;
+        
+        protected BaseSingleSourceLogic(SourceShape<TMessage> shape, TaskCompletionSource<NotUsed> completion,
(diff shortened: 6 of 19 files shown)


IgorFedchenko:
Because of making lots of local testing and aborting tests, I really needed this PR to be merged: #47 . So, I pulled changes from there to my working branch.

Aaronontheweb:
Merged #47 and updated this PR with latest.

src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
```diff
@@ -98,7 +99,7 @@ public CommittableOffset(IInternalCommitter committer, PartitionOffset offset, s
         /// </summary>
         public Task Commit()
         {
-            return _committer.Commit();
+            return _committer.Commit(new List<PartitionOffset>() { Offset }.ToImmutableList());
```
Aaronontheweb: Just use `ImmutableList.Create` here

src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
```diff
@@ -35,16 +37,22 @@ public sealed class ConsumerSettings<TKey, TValue>
...
         public TimeSpan PollInterval { get; }
         public TimeSpan PollTimeout { get; }
+        public TimeSpan CommitTimeout { get; }
+        public TimeSpan StopTimeout { get; }
         public int BufferSize { get; }
         public string DispatcherId { get; }
         public IImmutableDictionary<string, string> Properties { get; }
 
-        public ConsumerSettings(IDeserializer<TKey> keyDeserializer, IDeserializer<TValue> valueDeserializer, TimeSpan pollInterval, TimeSpan pollTimeout, int bufferSize, string dispatcherId, IImmutableDictionary<string, string> properties)
+        public ConsumerSettings(IDeserializer<TKey> keyDeserializer, IDeserializer<TValue> valueDeserializer, TimeSpan pollInterval,
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -0,0 +1,159 @@
...
+                    TopicPartitions = TopicPartitions.Union(assignmentWithOffset.TopicPartitions.Select(tp => tp.TopicPartition));
+                    break;
+            }
+        }
+
+        private void MessageHandling(Tuple<IActorRef, object> args)
+        {
+            switch (args.Item2)
+            {
+                case KafkaConsumerActorMetadata.Internal.Messages<K, V> msg:
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -0,0 +1,159 @@
...
+                else if (!_requested && TopicPartitions.Any())
+                {
+                    RequestMessages();
+                }
+            }
+        }
+
+        protected void RequestMessages()
+        {
+            _requested = true;
```
Aaronontheweb: So the `_requested` field is mostly just for accounting purposes, to make sure we're not re-requesting data when the buffer is already filling up, right?

Aaronontheweb: Ditto with `_requestId` ?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -1,132 +1,146 @@
...
                     break;
+                default:
+                    throw new NotSupportedException();
             }
+        }
 
-            _partitionsAssigned = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsAssigned);
-            _partitionsRevoked = GetAsyncCallback<IEnumerable<TopicPartitionOffset>>(PartitionsRevoked);
+        /// <inheritdoc />
+        protected override IActorRef CreateConsumerActor()
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -1,132 +1,146 @@
...
-        // Consumer's events
-        //
+        protected override void PerformShutdown()
+        {
+            SetKeepGoing(true);
+            
+            if (!IsClosed(_shape.Outlet))
+                Complete(_shape.Outlet);
+            
+            SourceActor.Become(ShuttingDownReceive);
```
Aaronontheweb: :+1:

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -0,0 +1,426 @@
...
+            _log = Context.GetLogger();
+        }
+
+        protected override bool Receive(object message)
+        {
+            switch (message)
+            {
+                case KafkaConsumerActorMetadata.Internal.Assign assign:
+                {
+                    _log.Debug($"Got assign request from {Sender}");
```
Aaronontheweb: One performance nitpick - better off using the old school string formatting syntax than string interpolation, because due to the way Akka.NET's logging is designed we can actually defer allocation of strings until we're ready to print the log statement out to console using the old technique. If the LogLevel is set at INFO, which is is by default, this saves us an unnecessary allocation each time one of these `Assign` messages is processed.

Not so with the new one. I'm not sure how "high traffic" this section of actor code could be, but that's something to bear in mind. It can really add up in busy ActorSystems.

src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        public static string JoinToString<T>(this IEnumerable<T> collection, string separator)
+        {
+            return string.Join(separator, collection);
+        }
+
+        /// <summary>
+        /// Converts dictionary to a list of tuples
+        /// </summary>
+        public static IEnumerable<(TKey, TValue)> ToTuples<TKey, TValue>(this IEnumerable<KeyValuePair<TKey, TValue>> dictionary)
```
Aaronontheweb: For the benefit of other OSS contributors: `ValueTuple` is a struct so this isn't really an "allocation" worth fretting about.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -1,132 +1,146 @@
...
         {
-            _partitionsAssigned(list);
+            switch (args.Item2)
+            {
+                case Terminated terminated when terminated.ActorRef.Equals(ConsumerActor):
+                    OnShutdown();
+                    CompleteStage();
+                    break;
+                default:
+                    return;
```
Aaronontheweb: So we ignore all messages until the `ConsumerActor` is terminated?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -0,0 +1,426 @@
...
+        }
+
+        private void SchedulePoolTask()
+        {
+            _poolCancellation?.Cancel(); // Stop existing scheduling, if any
+            
+            _poolCancellation = Context.System.Scheduler.ScheduleTellOnceCancelable(_settings.PollInterval, Self, _pollMessage, Self);
+        }
+
+        private void CheckOverlappingRequests(string updateType, IActorRef fromStage, IImmutableSet<TopicPartition> topics)
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Exceptions/StoppingException.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Akka.Streams.Kafka.Stages.Consumers.Exceptions
+{
+    /// <summary>
+    /// Thrown in response to commit/message request commands by consuming actor when in stopping state
+    /// </summary>
+    public class StoppingException : Exception
```
Aaronontheweb: Any reason why we shouldn't mark this class as `sealed` ?

NuGet.Config
```diff
@@ -6,5 +6,6 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="Akka.NET Nightly" value="https://www.myget.org/F/akkadotnet/api/v2" />
```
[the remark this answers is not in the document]
IgorFedchenko: Maybe we will keep this reference for this PR and will update it once beta2 will be released in separate PR

src/Akka.Streams.Kafka/Akka.Streams.Kafka.csproj
```diff
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Akka.Streams" Version="$(AkkaVersion)" />
+    <PackageReference Include="Akka.Streams" Version="1.4.0-beta637012445454141588" />
```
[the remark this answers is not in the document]
IgorFedchenko: Ah, you are absolutely right, did not notice this update.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
IgorFedchenko: Never thought about that... This needs to be reviewed in a whole code base of this project, because I am using string interpolation whenever possible... Created separate issue for this #49 to do later

src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
IgorFedchenko: Sorry, did not understand what change to you propose. I used this extension to be able to write code like that:
```csharp
foreach (var (first, second) in dictionary.ToTuples()) { }
```
Because `KeyValuePair` does not have `Deconstruct` method to be converted to `ValueTuple `automatically, but in this project such conversion is quite handy.

src/Akka.Streams.Kafka/Stages/Consumers/Exceptions/StoppingException.cs
IgorFedchenko: Yep, you are right.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
IgorFedchenko: Yes, this is what alpakka does.
I think this is mostly because we get to this actor state only after getting `onDownstreamFinished` state event, so no more point of using this messages anyway.

NuGet.Config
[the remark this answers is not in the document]
IgorFedchenko: Created issue #50 for this

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -103,11 +103,11 @@ await Source
 
             var (task, probe1) = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
                 .WhereNot(c => c.Record.Value == InitialMsg)
-                .SelectAsync(10, elem =>
+                .SelectAsync(10, async elem =>
                 {
-                    elem.CommitableOffset.Commit();
+                    await elem.CommitableOffset.Commit();
```
IgorFedchenko: Commit returns task now instead of been blocking

src/Akka.Streams.Kafka/Helpers/CommitRefreshing.cs
```diff
@@ -0,0 +1,118 @@
...
+        void Add(IImmutableSet<TopicPartitionOffset> offsets);
+        void Committed(IImmutableSet<TopicPartitionOffset> offsets);
+        void Revoke(IImmutableSet<TopicPartition> revokedTopicPartitions);
+        IImmutableSet<TopicPartitionOffset> RefreshOffsets { get; }
+        void UpdateRefreshDeadlines(IImmutableSet<TopicPartition> topicPartitions);
+        void AssignedPositions(IImmutableSet<TopicPartition> assignedPartitions, IImmutableSet<TopicPartitionOffset> assignedOffsets);
+        void AssignedPositions(IImmutableSet<TopicPartition> assignedTopicPartitions, IConsumer<K, V> consumer, TimeSpan positionTimeout);
+    }
+
+    internal static class CommitRefreshing
```
IgorFedchenko: Seems like this class is used for "commit refresh" feature: storing committed offsets to re-commit them if no new commits were made for some time (set by `commit-refresh-interval` setting and can be disabled).

src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+using Akka.Streams.Stage;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Helpers
+{
+    /// <summary>
+    /// This interface is used to pass callbacks when Kafka rebalances partitions between consumers,
+    /// or an kafka consumer is stopped
+    /// </summary>
+    internal interface IPartitionEventHandler
```
IgorFedchenko: This interface is public in alpakka implementation, but making it public will require passing it's implementations as an arguments in different places, adding more checks etc - so left it internal for now.
It is already used by stages to provide their callback for partition events, but making it public in future will allow library users to subscribe with their own partition handlers (or kind of this).
Also worth to mention that this interface is marked with `@ApiMayChanged` attribute in scala, so maybe even better to keep it internal for now.

src/Akka.Streams.Kafka/Helpers/RestrictedConsumer.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Helpers
+{
+    /// <summary>
+    /// Offers parts of <see cref="IConsumer{K,V}"/> API which becomes available to <see cref="IPartitionEventHandler"/>
+    /// </summary>
+    public class RestrictedConsumer<K, V>
```
IgorFedchenko: This class is not used right now and exists to be used with `IPartitionEventHandler` not-yet-public interface to provide library users some safe ways to interact with kafka consumer.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
IgorFedchenko: Yes. And `_requestId` is used to make sure that we will wait for the last response from consumer actor before will reset `_requested` flag and request more messages for empty buffer again.

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
```diff
@@ -12,24 +20,45 @@ internal interface IInternalCommitter
...
-        private readonly IConsumer<K, V> _consumer;
+        private readonly TimeSpan _commitTimeout;
+        private readonly Lazy<IActorRef> _consumerActor;
 
-        public KafkaCommitter(IConsumer<K, V> consumer)
+        public KafkaAsyncConsumerCommitter(Func<IActorRef> consumerActorFactory, TimeSpan commitTimeout)
         {
-            _consumer = consumer;
+            _commitTimeout = commitTimeout;
+            _consumerActor = new Lazy<IActorRef>(consumerActorFactory);
```
IgorFedchenko: Need deferred execution here because consumer actor is not instantiated at the moment of `KafkaAsyncConsumerCommitter` creation

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -0,0 +1,513 @@
...
+                    var message = _consumer.Consume(TimeSpan.FromMilliseconds(1));
+                    if (message != null)
+                        throw new IllegalActorStateException("Got unexpected Kafka message");
+                }
+                else
+                {
+                    // resume partitions to fetch
+                    IImmutableSet<TopicPartition> partitionsToFetch = _requests.Values.SelectMany(v => v.Topics).ToImmutableHashSet();
+                    /*
+                    // TODO: Make consumer pause-resume calls work
```
IgorFedchenko: @Aaronontheweb This is a most tricky thing left in this actor.
For some reason, if I will uncomment

IgorFedchenko: @Aaronontheweb This is a most tricky thing left in this actor.
For some reason, if I will uncomment `_consumer.Resume()` call the following `_consumer.Consume()` will reset consumer's offset to 0, returning the first message in the partition. If we do not use pause-resume stuff, all works fine - at least until we will need to switch between different partitions, which is not the case so far but will be required when actor is shared between different stages.
I have found an [issue](https://github.com/confluentinc/confluent-kafka-dotnet/issues/672) with confluent driver, but it is marked as resolved for almost a year. I spent some time on it, and decided to get back as a separate issue/PR.

So far, I have commented out all pause-resume lines from implementation, so instead of initially pausing assigned partitions and resuming them on-demand, consumer actor is now consuming all assigned partitions all the time. If used will changing partitions set, this can lead to skipping messages from partitions, that do not currently have request for (such partitions will be just consumed and results will be ignored).

IgorFedchenko:
I was going to submit this PR for review, but one of the tests failing in CI while running well on my local machine. I will put few more commits to get more informative logs about this error, and make it pass.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
IgorFedchenko: All right, after fixing CI tests failing I have realized that we do not require pause-resume offset that much: it is possible to reset consumer's position each time we consume message from wrong partition. I updated implementation, so now it basically should work even with different partition sets. Sure, not consuming would be better then consuming and putting offset back, but... That is what I have right now.

Aaronontheweb:
@IgorFedchenko looking at it now

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: @IgorFedchenko Can you create an issue for the Confluent driver if you think it's a bug with their code? Would love to see that cross-referenced somewhere on this repository too.

src/Akka.Streams.Kafka/Stages/Consumers/Committers.cs
Aaronontheweb: Makes sense

src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
Aaronontheweb: You should be able to use the `ApiMayChange` attribute from Akka - that's implemented in the project:

https://github.com/akkadotnet/akka.net/blob/41c116d2a3d280481e584401608ec848dc147632/src/core/Akka/Annotations/Attributes.cs - go ahead and add that where it's needed.

src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
Aaronontheweb: I wasn't proposing any changes - your code is fine as-is. Was just stating for other members of our OSS org that this doesn't represent any unnecessary allocations since `ValueTuple`s are on the stack. Unnecessary allocations are a type of memory issue we look out for especially when working with collections.

src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
Aaronontheweb: See my other comments about `ApiMayChange` in Akka.NET

src/Akka.Streams.Kafka/Messages/CommittableMessage.cs
```diff
@@ -98,7 +99,7 @@ public CommittableOffset(IInternalCommitter committer, PartitionOffset offset, s
         /// </summary>
         public Task Commit()
         {
-            return _committer.Commit();
+            return _committer.Commit(ImmutableList.Create(Offset));
```
Aaronontheweb: Why this change?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Aaronontheweb: Got it, makes sense

src/Akka.Streams.Kafka/Stages/Consumers/Exceptions/CommitTimeoutException.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Akka.Streams.Kafka.Stages.Consumers.Exceptions
+{
+    /// <summary>
+    /// Calls to <see cref="IInternalCommitter.Commit"/> will be failed with this exception if
```
Aaronontheweb: Appreciate the explanation - thank you.

PR #52, opened 2019-09-13
Added option to customize partition events by stage logic subclasses

IgorFedchenko:
This is a small PR, which adds a way to customize partition events handling by different stage logic classes.

Before this PR partitions events were handled by `SingleSourceStageLogic` without any option to override it. Besides, original kafka driver callbacks also provide `consumer` instance passed to callbacks, so actually handling could use some consumer's methods.

Now consumer is passed to this callbacks, and the callbacks itself may be customized by logic subclasses (looks like only for `TransactionalSourceLogic` so far).

### DIFF
--- a/src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
+++ b/src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
@@ -7,53 +7,59 @@ using Confluent.Kafka;
 namespace Akka.Streams.Kafka.Helpers
 {
     /// <summary>
-    /// This interface is used to pass callbacks when Kafka rebalances partitions between consumers,
-    /// or an kafka consumer is stopped
+    /// The API is new and may change in further releases.
+    ///
+    /// Allows to execute user code when Kafka rebalances partitions between consumers, or an Alpakka Kafka consumer is stopped.
+    /// Use with care: These callbacks are called synchronously on the same thread Kafka's `poll()` is called.
+    /// A warning will be logged if a callback takes longer than the configured `partition-handler-warning`.
+    ///
+    /// There is no point in calling `CommittableOffset`'s commit methods as their committing won't be executed as long as any of
+    /// the callbacks in this class are called.
     /// </summary>
     [ApiMayChange]
-    internal interface IPartitionEventHandler
+    internal interface IPartitionEventHandler<K, V>
     {
         /// <summary>
         /// Called when partitions are revoked
         /// </summary>
-        void OnRevoke(IImmutableSet<TopicPartitionOffset> revokedTopicPartitions);
+        void OnRevoke(IImmutableSet<TopicPartitionOffset> revokedTopicPartitions, RestrictedConsumer<K, V> consumer);
 
         /// <summary>
         /// Called when partitions are assigned
         /// </summary>
-        void OnAssign(IImmutableSet<TopicPartition> assignedTopicPartitions);
+        void OnAssign(IImmutableSet<TopicPartition> assignedTopicPartitions, RestrictedConsumer<K, V> consumer);
 
         /// <summary>
         /// Called when consuming is stopped
         /// </summary>
-        void OnStop(IImmutableSet<TopicPartition> topicPartitions);
+        void OnStop(IImmutableSet<TopicPartition> topicPartitions, RestrictedConsumer<K, V> consumer);
     }
 
     /// <summary>
-    /// Dummy handler which does nothing. Also <see cref="IPartitionEventHandler"/>
+    /// Dummy handler which does nothing. Also <see cref="IPartitionEventHandler{K,V}"/>
     /// </summary>
-    internal class EmptyPartitionEventHandler : IPartitionEventHandler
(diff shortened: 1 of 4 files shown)


IgorFedchenko:
During PR #48 review process, I thought that `IPartitionEventHandler` interface will be made public. But currently all references in original alpakka project, that use this interface, are private/internal. So actually making it public will not give library users any option of use this interface - so keeping it internal so far. We can switch it to public whenever there will be any option to use it from outside.

src/Akka.Streams.Kafka/Helpers/PartitionEventHandler.cs
```diff
@@ -7,53 +7,59 @@
...
+    /// Allows to execute user code when Kafka rebalances partitions between consumers, or an Alpakka Kafka consumer is stopped.
+    /// Use with care: These callbacks are called synchronously on the same thread Kafka's `poll()` is called.
+    /// A warning will be logged if a callback takes longer than the configured `partition-handler-warning`.
+    ///
+    /// There is no point in calling `CommittableOffset`'s commit methods as their committing won't be executed as long as any of
+    /// the callbacks in this class are called.
     /// </summary>
     [ApiMayChange]
-    internal interface IPartitionEventHandler
+    internal interface IPartitionEventHandler<K, V>
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SingleSourceStageLogic.cs
```diff
@@ -61,7 +61,9 @@ protected override IActorRef CreateConsumerActor()
             var partitionsAssignedHandler = GetAsyncCallback<IEnumerable<TopicPartition>>(PartitionsAssigned);
             var partitionsRevokedHandler = GetAsyncCallback<IEnumerable<TopicPartitionOffset>>(PartitionsRevoked);
 
-            var eventHandler = new AsyncCallbacksPartitionEventHandler(partitionsAssignedHandler, partitionsRevokedHandler);
+            IPartitionEventHandler<K, V> eventHandler = new AsyncCallbacksPartitionEventHandler<K,V>(partitionsAssignedHandler, partitionsRevokedHandler);
+            // This allows to override partition events handling by subclasses
+            eventHandler = AddToPartitionAssignmentHandler(eventHandler);
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -499,15 +499,20 @@ public override void OnPartitionsAssigned(IImmutableSet<TopicPartition> partitio
...
+                
                 _actor._rebalanceInProgress = false;
             }
 
             public override void OnPartitionsRevoked(IImmutableSet<TopicPartitionOffset> partitions)
             {
-                // TODO: Add warning if IPartinionEventHandler will be public and call takes more then _warningDuration
-                _partitionEventHandler.OnRevoke(partitions);
+                var watch = Stopwatch.StartNew();
+                _partitionEventHandler.OnRevoke(partitions, _restrictedConsumer);
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -499,15 +499,20 @@ public override void OnPartitionsAssigned(IImmutableSet<TopicPartition> partitio
...
+                CheckDuration(watch, "onRevoke");
+
```
Aaronontheweb: @IgorFedchenko do we need to `.Dispoe()` `Stopwatch` here?

IgorFedchenko: It does not implement `IDisposable` interface, but I do `watch.Stop()` in `CheckDuration` method implementation. 
But while technically this is correct, this looks unobvious - I thought about that when was writing this, but decided to save coding lines :) Added explicit `Stop()` call before passing `watch` to `CheckDuration` method, now should look better.

Aaronontheweb: Ah got it, thanks for explaining

Aaronontheweb:
@IgorFedchenko FYI, on the previous commit had one test failure:

Akka.Streams.Kafka.Tests.Akka.Streams.Kafka.Tests.Integration.PlainSourceIntegrationTests.PlainSource_consumes_messages_from_KafkaProducer_with_subscribe_to_topic

You can see the logs and such if you click through here: https://github.com/akkadotnet/Akka.Streams.Kafka/runs/227053848

Might have a race condition of some kind with that test.

IgorFedchenko:
@Aaronontheweb Created [issue](https://github.com/akkadotnet/Akka.Streams.Kafka/issues/57) for that - looks little bit wierd, so unfortunately can not fix this right now, "on the fly".

PR #53, opened 2019-09-16
Added option to reuse existing container

IgorFedchenko:
While starting new containers and performing resources cleanup after each tests run is useful for CI, this takes time during local development, when you need to start-stop same tests multiple time (for debugging, etc).

This small PR adds possibility to use `docker-compose up` from project root folder to start kafka container, and then reuse this container for all further test runs. 

It uses environment variables as a configuration source to avoid changing anything between local testing and pushing updates for CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,3 +118,27 @@ KafkaConsumer.CommitableSource(consumerSettings, Subscriptions.Topics("topic1"))
 The above example uses separate mapAsync stages for processing and committing. This guarantees that for parallelism higher than 1 we will keep correct ordering of messages sent for commit.
 
 Committing the offset for each message as illustrated above is rather slow. It is recommended to batch the commits for better throughput, with the trade-off that more messages may be re-delivered in case of failures.
+
+## Local development
+
+There are some helpers to simplify local development
+
+### Tests: File logging
+
+Sometimes it is useful to have all logs written to a file in addition to console. 
+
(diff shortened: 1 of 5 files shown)


src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -44,13 +44,24 @@ public KafkaFixture()
                 throw new NotSupportedException($"Unsupported OS [{RuntimeInformation.OSDescription}]");
 
             _client = config.CreateClient();
+
+            if (TestsConfiguration.UseExistingDockerContainer)
```
Aaronontheweb: LGTM

IgorFedchenko:
Indeed... I have updated readme, so now it should be easier for new developers to use this environment tricks. Let me know if it looks fine and clear

README.md
```diff
@@ -117,4 +117,28 @@ KafkaConsumer.CommitableSource(consumerSettings, Subscriptions.Topics("topic1"))
 ```
 The above example uses separate mapAsync stages for processing and committing. This guarantees that for parallelism higher than 1 we will keep correct ordering of messages sent for commit.
 
-Committing the offset for each message as illustrated above is rather slow. It is recommended to batch the commits for better throughput, with the trade-off that more messages may be re-delivered in case of failures.
\ No newline at end of file
+Committing the offset for each message as illustrated above is rather slow. It is recommended to batch the commits for better throughput, with the trade-off that more messages may be re-delivered in case of failures.
+
+## Local development
```
Aaronontheweb: This looks great, thank you.

PR #54, opened 2019-09-16
Pause/Resume partitions on demand

### LINKED ISSUE #51: Make use of Pause()/Resume() kafka consumer's methods in consuming actor
Right now we are consuming all assigned partitions all the time, and when consumed message from wrong partition (i.e. none of the consuming sources currently waiting messages from them) we use `Seek()` consumer's method to rollback consumes offset.

We need to pause those partitions and resume once there is a demand on it.

IgorFedchenko:
This is related to issue #51

Turned out, that while confluent driver is possibly doing wrong when resetting offset when resuming already resumed partitions, there is a simple workaround of storing already resumed partitions list. 

See more details in [this comment](https://github.com/akkadotnet/Akka.Streams.Kafka/issues/51#issuecomment-531761987)

Close #51

### DIFF
--- a/src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
+++ b/src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
@@ -49,6 +49,7 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Actors
         private bool _stopInProgress = false;
         private bool _delayedPoolInFlight = false;
         private RebalanceListenerBase _partitionAssignmentHandler = new EmptyRebalanceListener();
+        private IImmutableSet<TopicPartition> _resumedPartitions = ImmutableHashSet<TopicPartition>.Empty;
 
         /// <summary>
         /// While `true`, committing is delayed.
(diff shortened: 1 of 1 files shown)


src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -285,28 +286,22 @@ private void Poll()
...
                     IImmutableSet<TopicPartition> partitionsToFetch = _requests.Values.SelectMany(v => v.Topics).ToImmutableHashSet();
-                    /*
-                    // TODO: Make consumer pause-resume calls work
                     var resumeThese = currentAssignment.Where(partitionsToFetch.Contains).ToList();
                     var pauseThese = currentAssignment.Except(resumeThese).ToList();
                     _consumer.Pause(pauseThese);
-                    _consumer.Resume(resumeThese);
-                    */
+                    _consumer.Resume(resumeThese.Except(_resumedPartitions));
+                    _resumedPartitions = _resumedPartitions.Union(resumeThese);
```
Aaronontheweb: @IgorFedchenko do we have any specs at the moment which cover pause / resume behavior? Given the issue with the underlying Confluent driver, we should have a regression check in-place just to ensure that if they patch the driver in a future release it doesn't break the work-around we've added in here. Is that something you can add or does it exist already?

IgorFedchenko: @Aaronontheweb First of all, before this workaround current tests were failing (that is why all pause/resume code was commented out) - so if they will break it that much tests will fail again.
But I understand what are you talking about, seems like we need some special tests to ensure pause/resume partitions working correctly.
While current tests make use of this pause/resume calls, currently there is never any partitions to pause actually - because all subscribed partitions are the ones that stage is interested in. The real pause/resume usage will start when I will add `ExternalSingleSource` stage with `ExternalSingleSourceLogic` class - because that kind of stage allows sharing this consuming actor between stages. And that is where multiple stages will requests for different partitions consumed, and some of them will be paused (when there is no demand on them) - and some of them should be resumed/consumed.
So that is great idea to make sure we have some regression tests here - I've just created issue #55 to not forget about that. I think I will close it once implement that `ExternalSingleSource` with it's tests.

Aaronontheweb: Ok, sounds good - I'll merge this in for now then and we can add a future PR to address that going forward.

PR #56, opened 2019-09-18
External plain source

### LINKED ISSUE #55: Add test coverage for Pause/Resume partitions scenario
Seems like real pause/resume usage will start when `ExternalSingleSource` stage with `ExternalSingleSourceLogic` class will be implemented - so need to make sure that while adding tests for it we will get pause/resume scenario covered with tests. At least to make sure any confluent driver update will not break anything here.

IgorFedchenko:
One more step to resolve issue #36 - this PR adds ExternalPlainSource stage.

Basically, this stage allows to make use of some existing consuming actor, and get messages from him instead of creating it's own consuming actor.

Implementation is pretty simple, but while adding test coverage we'll need to make sure that `Pause()`\`Resume()` scenario is tested (issue #55 )

Close #55 

It is work in progress for now (need more tests), that's why making it a draft PR.


IgorFedchenko:
@Aaronontheweb Please keep in mind that this PR is based on #53 and #54 PRs, so we would better have them merged first.

Aaronontheweb:
@IgorFedchenko will do - letting CI run for those now.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/ExternalSingleSourceLogic.cs
```diff
@@ -1,7 +1,38 @@
...
+                                         TaskCompletionSource<NotUsed> completion, Attributes attributes, 
+                                         Func<BaseSingleSourceLogic<K, V, TMessage>, IMessageBuilder<K, V, TMessage>> messageBuilderFactory) 
+            : base(shape, completion, attributes, messageBuilderFactory)
+        {
+            _consumerActor = consumerActor;
+            _subscription = subscription;
+        }
+
+        /// <inheritdoc />
+        protected override IActorRef CreateConsumerActor() => _consumerActor;
```
Aaronontheweb: I'm sure I'm missing something, but where does the output for this get consumed?

IgorFedchenko: No problem.
This `ExternalSingleSourceLogic` class is derived from `BaseSingleSourceLogic`, which contains all message consuming behavior. Child classes just specify, how to get consumer actor reference (usually create own actor or reuse existing, like here), and then base class requests messages from this actor and handles his responses.
Is that what you were asking about?

Aaronontheweb: Ah yes, thanks - just wanted to have the big picture for when I review this later. Seems simple enough.

IgorFedchenko:
While adding and verifying pause/resume test, I discovered one more pretty serious [issue](https://github.com/akkadotnet/Akka.Streams.Kafka/issues/58) which may lead to message loss when stages subscribed to multiple partitions. 
I want to move this issue fix to separate PR, because technically this PR works and implements `ExternalPlainSource` stage well. And there is also a separate test to check pause/resume behavior, which also passes now.
So I will put some comments over changes here and move this PR to review.

docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
       - zookeeper
     environment:
       KAFKA_BROKER_ID: 1
+      KAFKA_NUM_PARTITIONS: 3
```
IgorFedchenko: This is required now, because default partition count is 1 and this is not sufficient for most test cases, where subscriptions to multiple partitions are required.
And that is why most of tests are updated to publish messages to particular partition instead of topic - because kafka will load-balance messages between partitions and tests, consuming only one (first) partition would be failing.

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -18,41 +18,9 @@ namespace Akka.Streams.Kafka.Tests.Integration
 {
     public class CommittableSourceIntegrationTests : KafkaIntegrationTests
     {
-        private const string InitialMsg = "initial msg in topic, required to create the topic before any consumer subscribes to it";
```
IgorFedchenko: All common setup is moved to base `KafkaIntegrationTests` class

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
```diff
@@ -61,21 +29,22 @@ public async Task CommitableSource_consumes_messages_from_Producer_without_commi
             int elementsCount = 100;
             var topic1 = CreateTopic(1);
             var group1 = CreateGroup(1);
+            var topicPartition1 = new TopicPartition(topic1, 0);
 
-            await GivenInitializedTopic(topic1);
+            await GivenInitializedTopic(topicPartition1);
```
IgorFedchenko: See my notes about changing number of partitions and using concrete partition instead of common topic name.
There are lots of changes like that, so I will comment about this only once.

src/Akka.Streams.Kafka.Tests/Integration/ExternalPlainSourceIntegrationTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            
+            // Make sure source tasks finish accordingly
+            AwaitCondition(() => partitionTask1.IsFaulted && partitionTask2.IsFaulted && partitionTask3.IsCompletedSuccessfully);
+            
+            // Cleanup
+            consumer.Tell(new KafkaConsumerActorMetadata.Internal.Stop(), ActorRefs.NoSender);
+        }
+
+        [Fact]
+        public async Task ExternalPlainSource_verify_consuming_actor_pause_resume_partitions_works_fine()
```
IgorFedchenko: This test is specially implemented for issue #55 - there is no corresponding test in alpakka project.

src/Akka.Streams.Kafka.Tests/KafkaFixture.cs
```diff
@@ -82,6 +82,7 @@ await CreateContainer(ZookeeperImageName, ZookeeperImageTag, _zookeeperContainer
             await CreateContainer(KafkaImageName, KafkaImageTag, _kafkaContainerName, KafkaPort, new Dictionary<string, string>()
             {
                 ["KAFKA_BROKER_ID"] = "1",
+                ["KAFKA_NUM_PARTITIONS"] = "3",
```
IgorFedchenko: Need to update docker container configuration as well

src/Akka.Streams.Kafka.Tests/Integration/CommittableSourceIntegrationTests.cs
IgorFedchenko: Also, I realize that I could use separate test fixture for different tests (that allow 1 or require multiple partitions) - but this will not allow having single container from `docker-compose.yml` to be used for all tests, so I have updated tests to be ready for multiple partitions existance.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -345,6 +345,7 @@ private void ProcessResult(IImmutableSet<TopicPartition> partitionsToFetch, Cons
                 {
                     var messages = ImmutableList<ConsumeResult<K, V>>.Empty.Add(consumedMessage);
                     stageActorRef.Tell(new KafkaConsumerActorMetadata.Internal.Messages<K, V>(request.RequestId, messages));
+                    // _requests = _requests.Remove(stageActorRef);
```
IgorFedchenko: This is what I was talking about in issue #58 - will fix this in a separate PR, because need to rewrite request/response management between consuming actor and stages.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActorMetadata.cs
```diff
@@ -1,13 +1,19 @@
...
 
 namespace Akka.Streams.Kafka.Stages.Consumers.Actors
 {
-    internal class KafkaConsumerActorMetadata
+    /// <summary>
+    /// Containes metadata for <see cref="KafkaConsumerActor{K,V}"/>.
+    /// Generally this should not be used from outside of the library.
+    /// </summary>
+    [InternalApi]
+    public class KafkaConsumerActorMetadata
```
IgorFedchenko: Has to be public to be available in tests

IgorFedchenko:
@Aaronontheweb There is also one more PR #59 that is based on this one, so after merging this you can review that one. It addresses issue #58 which was discovered during testing.

docker-compose.yml
Aaronontheweb: Makes sense

src/Akka.Streams.Kafka.Tests/Integration/ExternalPlainSourceIntegrationTests.cs
Aaronontheweb: Great, thank you

src/Akka.Streams.Kafka.Tests/KafkaIntegrationTests.cs
```diff
@@ -1,19 +1,88 @@
...
 namespace Akka.Streams.Kafka.Tests
 {
     [Collection(KafkaSpecsFixture.Name)]
     public class KafkaIntegrationTests : Akka.TestKit.Xunit2.TestKit
     {
-        public KafkaIntegrationTests(string actorSystemName, ITestOutputHelper output) 
+        private readonly KafkaFixture _fixture;
+        protected IMaterializer Materializer { get; }
+
+        public KafkaIntegrationTests(string actorSystemName, ITestOutputHelper output, KafkaFixture fixture)
```
Aaronontheweb: Makes sense to move all of this boilerplate code to the `KafkaIntegrationTests` fixture so it can be re-used.

IgorFedchenko: @Aaronontheweb Well, `KafkaIntegrationTests` is a base class for all kafka tests by itself, so all of this code is going to be reused. Maybe would be nice to make it `abstract` by the way... Is it what you were talking about?

PR #59, opened 2019-09-19
Fix requests management

### LINKED ISSUE #58: Consumer actor does not perform proper incoming requests cleanup
While working on #56 , it turned out that consumer actor does not perform incoming requests cleanup when requests are served. This means that actor continues pooling records from all partitions that were requested since actor was started. And this may lead to message loss (messages are consumed, but no one stage is waiting for it).

Need to add test for that (that should initially fail) and then fix this.

Note: The tricky part here is that alpakka consumer is able to fetch multiple messages from multiple partitions at the same time (or at least multiple messages from one partition). But confluent driver allows consuming one message at a time. This difference makes impact on how we should request consuming actor for new messages and keep track of given responses. We can not just ask for "give me all messages from this list of partitions", and wait for all of them at once - like it is implemented right now.

IgorFedchenko:
This PR is going to close issue #58 
Close #58

It is based on PR #56 - so please review this only after merging that one. 

Turned out that because of buffering implemented in stage logic (when receiving messages from consumer actor) - no message loss will occur. Pending requests list management is used mostly to pause partitions without demand and stop further buffering. So this will just use more memory (and possible give `OutOfMemotyException` when running for a long time), and it is hard to write test to check that behavior.

But when trying to just remove answered requests from pending `_requests` list, existing tests were failing - turned out that `Pause()`\`Resume()` logic [workaround](https://github.com/akkadotnet/Akka.Streams.Kafka/issues/51#issuecomment-531761987) was not implemented well for cases when `_requests` are cleaning up properly, so fixed that in this PR too.


Aaronontheweb:
@IgorFedchenko looks like merging the previous PR introduced some conflicts

IgorFedchenko:
@Aaronontheweb Indeed - resolved them now.

PR #60, opened 2019-09-21
Kafka consumer AtMostOnceSource

IgorFedchenko:
This is really tiny PR as a part of work on issue #36, but may require some help.

Basically, `AtMostOnceSource` consuming source just commits all messages before passing them to downstream, that's why at-most-once delivery comes up.

But I am not completely sure I have translated this from scala to c# right.
Here is this how this code looks like in alpakka:
```scala
def atMostOnceSource[K, V](settings: ConsumerSettings[K, V],
                             subscription: Subscription): Source[ConsumerRecord[K, V], Control] =
    committableSource[K, V](settings, subscription).mapAsync(1) { m =>
      m.committableOffset.commitScaladsl().map(_ => m.record)(ExecutionContexts.sameThreadExecutionContext)
    }
```

My confusion is about `ExecutionContexts.sameThreadExecutionContext` option. Here is scala [docs](https://doc.akka.io/japi/akka/current/akka/dispatch/ExecutionContexts.sameThreadExecutionContext$.html) for it. Seems like this is a way to make task continuation execute on the same thread.

What I have used right now is:
```c#
return CommittableSource(settings, subscription).SelectAsync(1, async message =>
{
   await message.CommitableOffset.Commit();
   return message.Record;
});
```

But `async/await` makes continuation execute on the same thread context, not the same thread. While the code itself seems convenient and like something everybody could write to achieve documented behavior (and actually this way is already used in one of our tests), this looks like inaccurate translation from scala. 

Other option I see here is something like this:
```c#
return CommittableSource(settings, subscription).SelectAsync(1, message =>
{
   return message.CommitableOffset.Commit().ContinueWith(_ => message.Record, TaskContinuationOptions.ExecuteSynchronously);
});
```

This looks much more similar to what is written in scala, but I have never used `TaskContinuationOptions.ExecuteSynchronously` option so far and not sure this is something we need to use. Here is a [docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions?view=netframework-4.8) for it - should I use this way instead? @Aaronontheweb what do you think?

I have found some [akka.net issue](https://github.com/akkadotnet/akka.net/issues/147) that was also related to `ExecutionContexts.sameThreadExecutionContext` - but looks like guys decided to not implement this in akka.net packages. 

PR is still in progress, because I need to add tests for this. But I suspect that tests will pass with both implementations - so just need to make a decision about how to implement that.

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/AtMostOnceSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/AtMostOnceSourceIntegrationTests.cs
@@ -1,0 +1,70 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Streams.Dsl;
+using Akka.Streams.Kafka.Dsl;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.TestKit;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Kafka.Tests.Integration
+{
+    public class AtMostOnceSourceIntegrationTests : KafkaIntegrationTests
+    {
+        public AtMostOnceSourceIntegrationTests(ITestOutputHelper output, KafkaFixture fixture) 
+            : base(nameof(AtMostOnceSourceIntegrationTests), output, fixture)
+        {
+        }
+
+        [Fact]
+        public async Task AtMostOnceSource_Should_stop_consuming_actor_when_used_with_Take()
+        {
+            var topic = CreateTopic(1);
+            var group = CreateGroup(1);
+
+            await ProduceStrings(topic, Enumerable.Range(1, 10), ProducerSettings);
+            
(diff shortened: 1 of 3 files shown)


IgorFedchenko:
About few failing tests: actually, original scala test code does not set any timeout for the task (which was failing in our case) - they just get `futureValue` property and make assertion for it. 
But in case if something will go wrong with stage completion such test would be running forever (I have seen that picture before in this project) - so I prefer to set timeout, maybe let it bit enough, but make it fail if task is going to keep running forever.

IgorFedchenko:
@Aaronontheweb Actually, I am using async-await way in few other places like that in different PRs, no problems with tests etc, so maybe this could be the way to go.

Aaronontheweb:
> Actually, I am using async-await way in few other places like that in different PRs, no problems with tests etc, so maybe this could be the way to go.

yes, we want to move towards this model everywhere in the Akka.NET testkit eventually.

PR #62, opened 2019-09-26
SourceWithOffsetContext

IgorFedchenko:
As a part of work over issue #36 , `SourceWithOffsetContext` consumer source will be implemented and tested in this PR.

This souse is basically provides another way of passing consumed messages and committable offset to downstream. The main difference (as far as I understand) is that while `CommittableSource` passes `CommittableMessage` to the downstream, forcing consequent flows to handle both message and it's committable offset, this `SourceWithOffsetContext` allows to put committable offset to the source's context.

Also, some helpers like `Committer` static class with committing flows are implemented in this PR, because they are used to make testing simpler/cleaner. 

There is one more test for this source type, but it will require one more producer flow implementation, so it will be added in one of the next PRs.

### DIFF
--- a/src/Akka.Streams.Kafka/Dsl/KafkaProducer.cs
+++ b/src/Akka.Streams.Kafka/Dsl/KafkaProducer.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Messages;
 using Akka.Streams.Kafka.Settings;
@@ -65,6 +66,12 @@ namespace Akka.Streams.Kafka.Dsl
             return string.IsNullOrEmpty(settings.DispatcherId)
                 ? flow
                 : flow.WithAttributes(ActorAttributes.CreateDispatcher(settings.DispatcherId));
+        }
+
+        // TODO
+        public static FlowWithContext<IEnvelope<K, V, NotUsed>, C, DeliveryReport<K, V>, C, NotUsed> FlowWithContext<K, V, C>(ProducerSettings<K, V> settings)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/Akka.Streams.Kafka/Helpers/Committer.cs
+++ b/src/Akka.Streams.Kafka/Helpers/Committer.cs
@@ -1,0 +1,26 @@
+using Akka.Streams.Dsl;
+using Akka.Streams.Kafka.Messages;
+using Akka.Streams.Kafka.Settings;
+
+namespace Akka.Streams.Kafka.Helpers
+{
+    /// <summary>
+    /// Implements committing flows
+    /// </summary>
+    public static class Committer
+    {
+        /// <summary>
+        /// Batches offsets and commits them to Kafka, emits <see cref="CommittableOffsetBatch"/> for every committed batch.
+        /// </summary>
+        public static Flow<ICommittable, ICommittableOffsetBatch, NotUsed> BatchFlow(CommitterSettings settings)
+        {
+            return Flow.Create<ICommittable>().GroupedWithin(settings.MaxBatch, settings.MaxInterval)
+                .Select(CommittableOffsetBatch.Create)
+                .SelectAsync(settings.Parallelism, async batch =>
+                {
+                   await batch.Commit();
+                   return batch;
+                });
+        }
+    }
+}
--- a/src/Akka.Streams.Kafka/Helpers/Committer.cs
+++ b/src/Akka.Streams.Kafka/Helpers/Committer.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Threading.Tasks;
+using Akka.Annotations;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Messages;
 using Akka.Streams.Kafka.Settings;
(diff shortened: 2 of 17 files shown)


src/Akka.Streams.Kafka/Helpers/Committer.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <summary>
+        /// API MAY CHANGE
+        /// 
+        /// Batches offsets from context and commits them to Kafka, emits no useful value,
+        /// but keeps the committed <see cref="ICommittableOffsetBatch"/> as context
+        /// </summary>
+        [ApiMayChange]
+        public static FlowWithContext<ICommittableOffset, E, ICommittableOffsetBatch, NotUsed, NotUsed> FlowWithOffsetContext<E>(CommitterSettings settings)
+        {
+            var value = Akka.Streams.Dsl.Flow.Create<Tuple<E, ICommittableOffset>>()
```
IgorFedchenko: @Aaronontheweb I had to use `Tuple<K,V>` here because `FlowWithContext` class from Akka.Streams beta2 still uses it instead of `ValueTuple`. I remember that you wanted to move from `Tuple` class in new Akka.NET version, so is it something you are going to change in the future?

Aaronontheweb:
@IgorFedchenko have some merge conflicts here

IgorFedchenko:
Yes, each of my PRs adds new extension method to KafkaConsumer.cs file, so seems like we will have conflicts on each merge...

Aaronontheweb:
@IgorFedchenko ah, ok

src/Akka.Streams.Kafka/Dsl/KafkaProducer.cs
```diff
@@ -66,5 +67,11 @@ public static class KafkaProducer
                 ? flow
                 : flow.WithAttributes(ActorAttributes.CreateDispatcher(settings.DispatcherId));
         }
+
+        // TODO
+        public static FlowWithContext<IEnvelope<K, V, NotUsed>, C, DeliveryReport<K, V>, C, NotUsed> FlowWithContext<K, V, C>(ProducerSettings<K, V> settings)
```
Aaronontheweb: what do we need to do here?

Aaronontheweb: Anything from Akka.Streams itself?

src/Akka.Streams.Kafka/Helpers/Committer.cs
Aaronontheweb: Yeah, I need to update that in Akka.Streams

src/Akka.Streams.Kafka/Messages/Committable.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Threading.Tasks;
+
+namespace Akka.Streams.Kafka.Messages
+{
+   /// <summary>
+    /// Commit an offset that is included in a <see cref="CommittableMessage{K,V}"/>
+    /// If you need to store offsets in anything other than Kafka, this API
+    /// should not be used.
+    /// </summary>
+    public interface ICommittable
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Messages/CommittableOffset.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Akka.Streams.Kafka.Messages
+{
+    /// <summary>
+    /// Implementation of the offset, contained in <see cref="CommittableMessage{K,V}"/>.
+    /// Can be commited via <see cref="Commit"/> method.
+    /// </summary>
+    internal sealed class CommittableOffset : ICommittableOffsetMetadata
+    {
+        /// <inheritdoc />
+        public long BatchSize => 1;
```
Aaronontheweb: Batchsize can't be bigger than 1?

Aaronontheweb: Ah nevermind, I see why now.

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
```diff
@@ -0,0 +1,152 @@
...
+            return offsets.Aggregate(Empty, (batch, offset) => batch.Updated(offset));
+        }
+
+        /// <inheritdoc />
+        public async Task Commit()
+        {
+            if (Offsets.IsEmpty() || Committers.IsEmpty())
+                return;
+
+            await Committers.First().Value.Commit(this);
```
Aaronontheweb: Do all committers need to commit here? Why not?

src/Akka.Streams.Kafka/Messages/GroupTopicPartitionOffset.cs
```diff
@@ -0,0 +1,153 @@
+using System;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Messages
+{
+    /// <summary>
+    /// Offset position for a groupId, topic, partition.
+    /// </summary>
+    public sealed class GroupTopicPartitionOffset : IEquatable<GroupTopicPartitionOffset>
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Settings/CommitterSettings.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using Akka.Actor;
+using Akka.Configuration;
+
+namespace Akka.Streams.Kafka.Settings
+{
+    /// <summary>
+    /// Settings for committer. See 'akka.kafka.committer' section in reference.conf.
+    /// </summary>
+    public sealed class CommitterSettings
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/reference.conf
```diff
@@ -40,3 +40,15 @@ akka.kafka.default-dispatcher {
...
+# Committer flows use this settings to make batch commits
+akka.kafka.committer {
+    # Set max commit batch size
+    max-batch = 1
+    
+    # Set max commit interval
+    max-interval = 1s
+    
+    # Set level of parallelism
+    parallelism = 1
```
Aaronontheweb: Is this standard?

src/Akka.Streams.Kafka/Messages/CommittableOffset.cs
IgorFedchenko: Yes, just a special value for this offset implementation

src/Akka.Streams.Kafka/Dsl/KafkaProducer.cs
IgorFedchenko: I have defined this method just to put valid reference in xml comment for consumer's static method (because it is usually used in pair with this producer method). 
So this is just something I will implement in next PRs. Did not check out if something is missing from `Akka.Streams` package, but I think all is ok here.

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
IgorFedchenko: First of all, here is their code in alpakka:
```scala
if (offsets.isEmpty)
  Future.successful(Done)
else {
  committers.head._2.commit(this)
}
```

And why only one committer makes commit: that is because in committer's implementation it takes all `Committers` of this batch object, finds appropriate committer for each offsets group (you can see that `Committers `are grouped my groupId) and then commits them all. So internally multiple committers may be used. Here is a `Commit `implementation:

```csharp
public async Task Commit(ICommittableOffsetBatch batch)
{
	if (!(batch is CommittableOffsetBatch batchImpl))
		throw new ArgumentException($"Unknown CommittableOffsetBatch, got {batch.GetType().FullName}, but expected {nameof(CommittableOffsetBatch)}");
	
	await Task.WhenAll(batchImpl.OffsetsAndMetadata.GroupBy(o => o.Key.GroupId).Select(group =>
	{
		if (!batchImpl.Committers.TryGetValue(group.Key, out var committer))
			throw new IllegalStateException($"Unknown committer, got groupId = {group.Key}");

		var offsets = group.Select(offset => new GroupTopicPartitionOffset(offset.Key, offset.Value.Offset)).ToImmutableList();
		return committer.Commit(offsets);
	}));
}
```

Little bit messy, and the only place where this  `Commit(ICommittableOffsetBatch batch)` is used is that line you have a question about. So I could remove this method from committer and make committers iterating more explicit, right there. This will also allow make `Committers` private field, which seems convenient. What do you think?

src/Akka.Streams.Kafka/reference.conf
IgorFedchenko: Mm, no, my bad, did not think that I could check this values out from somewhere... But turned out that they are defined [here](https://doc.akka.io/docs/alpakka-kafka/current/consumer.html)
Updated right now

IgorFedchenko:
After updating committer default settings in `reference.conf` file one of the tests failed because I was expecting 10 messages on the downstream, but there are less of them due to `batchSize > 1` (so now expecting `elemsCount/batchSize` committable batches).

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
Aaronontheweb: This is a great response - no worries. This implementation is fine and answers my question.

PR #63, opened 2019-09-26
CommittableExternalSource

IgorFedchenko:
This PR is part of work over issue #36 and implements `CommittableExternalSource`.

This source is pretty simple and does the same thing as `PlainExternalSource`, but with committable messages - it uses some pre-existing consuming actor to consume/commit messages.

PR will be in draft stage until I will add tests for this new source.


IgorFedchenko:
@Aaronontheweb Hmm, turned out that alpakka project does not contain any tests for this source. This is understandable - this source is a composition of `PlainExternalSource` and `CommittableSource` and uses all the same classes, so basically if this two are working that `CommittableExternalSource` will work too.
Should I write my own test for this?

PR #71, opened 2019-10-05
IControl implementation

IgorFedchenko:
So far, all consumer stage's materialized value was `Task` - but actually it should be `IControl` object.

In addition to `IsShutdown` property, that returns task which finishes after stage is stopped, this object also has method for graceful stop of the source from outside. Which is very important, because kafka consuming stage will not stop by itself until downstream will be finished.

Tests are also updated, one of them uses new `Shutdown()` method to stop source instead of `probe.Cancel()` method that will finish downstream.

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using Akka.Configuration;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Dsl;
+using Akka.Streams.Kafka.Helpers;
 using Akka.Streams.Kafka.Messages;
 using Akka.Streams.Kafka.Settings;
 using Akka.Streams.Kafka.Tests.Logging;
@@ -28,13 +29,14 @@ namespace Akka.Streams.Kafka.Tests.Integration
         }
     
 
-        private TestSubscriber.Probe<string> CreateProbe(ConsumerSettings<Null, string> consumerSettings, ISubscription sub)
+        private Tuple<IControl, TestSubscriber.Probe<string>> CreateProbe(ConsumerSettings<Null, string> consumerSettings, ISubscription sub)
         {
             return KafkaConsumer
                 .PlainSource(consumerSettings, sub)
                 .Where(c => !c.Value.Equals(InitialMsg))
                 .Select(c => c.Value)
-                .RunWith(this.SinkProbe<string>(), Materializer);
+                .ToMaterialized(this.SinkProbe<string>(), Keep.Both)
+                .Run(Materializer);
         }
 
         [Fact]
@@ -51,7 +53,7 @@ namespace Akka.Streams.Kafka.Tests.Integration
 
             var consumerSettings = CreateConsumerSettings<string>(group1);
 
-            var probe = CreateProbe(consumerSettings, Subscriptions.Assignment(topicPartition1));
+            var (_, probe) = CreateProbe(consumerSettings, Subscriptions.Assignment(topicPartition1));
             
             probe.Request(elementsCount);
             foreach (var i in Enumerable.Range(1, elementsCount).Select(c => c.ToString()))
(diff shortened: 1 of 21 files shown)


src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
```diff
@@ -28,13 +29,14 @@ public PlainSourceIntegrationTests(ITestOutputHelper output, KafkaFixture fixtur
         }
     
 
-        private TestSubscriber.Probe<string> CreateProbe(ConsumerSettings<Null, string> consumerSettings, ISubscription sub)
+        private Tuple<IControl, TestSubscriber.Probe<string>> CreateProbe(ConsumerSettings<Null, string> consumerSettings, ISubscription sub)
```
Aaronontheweb: LGTM - I'm sure we'll switch this to `ValueTuple` once we've made that change on Akka.Streams itself in the next v1.4.0-beta release.

src/Akka.Streams.Kafka/Helpers/Control.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Threading.Tasks;
+using Akka.Streams.Kafka.Extensions;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Helpers
+{
+    /// <summary>
+    /// Materialized value of the consumer `Source`.
+    /// </summary>
+    public interface IControl
```
Aaronontheweb: LGTM

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -0,0 +1,88 @@
...
+using Akka.Streams.Kafka.Extensions;
+using Akka.Streams.Util;
+
+namespace Akka.Streams.Kafka.Helpers
+{
+    /// <summary>
+    /// Used in source logic classes to provide <see cref="IControl"/> implementation.
+    /// </summary>
+    /// <typeparam name="TSourceOut"></typeparam>
+    class PromiseControl<TSourceOut> : IControl
```
Aaronontheweb: Need to add access modifiers here.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -6,45 +6,56 @@
...
+        /// </summary>
+        protected readonly PromiseControl<TMessage> InternalControl;
         
         protected StageActor SourceActor { get; private set; }
         internal IActorRef ConsumerActor { get; private set; }
+
+        /// <summary>
+        /// Implements <see cref="IControl"/> to provide control over executed source
+        /// </summary>
+        public virtual IControl Control => InternalControl;
```
Aaronontheweb: Why have this be `virtual` but `InternalControl` be `readonly`? Is the idea that you can compose different `IControl` instances on top of each other?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -134,14 +132,13 @@ private void MessageHandling(Tuple<IActorRef, object> args)
...
                             // Throw
-                            _completion.TrySetException(exception);
                             FailStage(exception);
                             break;
                         case Directive.Resume:
                             // keep going
                             break;
                         case Directive.Restart:
-                            // keep going
+                            // TODO: Need to do something here: https://github.com/akkadotnet/Akka.Streams.Kafka/issues/33
```
Aaronontheweb: :+1: - no need to tackle this on this PR, but good to keep note of it.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
IgorFedchenko: In scala, every stage can override `PromiseControl`'s methods - so every stage can have it own implementation of `IControl` (usually overriding `PromiseControl`).
So `Control` is virtual, because we need to be able to set another `PromiseControl` inherited implementation as a control materialized value. And `InternalControl` is used internally because sometimes we need to call some `PromiseControl` methods, but stage exposes only `IControl`.

But you are right, actually - in scala stage uses mixins, but in general it exposes `PromiseControl` instance, not generic `IControl`... This will simplify things. I will update PR.

PR #65, opened 2019-09-28
PlainPartitionedSource

IgorFedchenko:
This PR is a part of work on issue #36 , and it implements `PlainPartitionedSource`.

This source allows to group consumed messages  by topic partition and provides stream of messages-per-partition streams.

This requires separate `SubSourceLogic` and `SubSourceStage` to be implemented (does not rely on existing `BaseSingleSourceLogic` base classes used before), but still uses consumer actor to work with kafka's `IConsumer` instance.

Issue #57, opened 2019-09-18
Got unexpected Kafka message error


IgorFedchenko:
As stated in [this comment](https://github.com/akkadotnet/Akka.Streams.Kafka/pull/52#issuecomment-532741378), one of the tests (possibly not single one) can fail with `IllegalActorStateException`.

The code that throws it is:
```csharp
var currentAssignment = _consumer.Assignment;
var initialRebalanceInProcess = _rebalanceInProgress;


try
{
    if (_requests.IsEmpty())
    {
        // no outstanding requests so we don't expect any messages back, but we should anyway
        // drive the KafkaConsumer by polling to handle partition events etc.
        _consumer.Pause(currentAssignment);
        var message = _consumer.Consume(TimeSpan.FromMilliseconds(1));
        if (message != null)
        {
            throw new IllegalActorStateException($"Got unexpected Kafka message: {message.ToJson()}");
        }
    }
```

So basically, we are pausing all currently assigned partitions and then consuming some service messages (like partition assigned events, etc) from kafka. But, in the original [alpakka code](url), they do this check - as well as we do - for some reason. 

This is not clear right now, but seems like this error is possible if we have partitions reassigned/resumed in separate thread and right after `Pause()` call some partition get resumed. But all `Resume()` calls are executed inside actor's message handling, so race conditions should be excluded... Needs to be checked and resolved.

IgorFedchenko:
@Aaronontheweb Do you think I need to spend time on this right now? On the one hand, this may be occasionally fixed with next PRs (some issues I have already found, but they seem unrelated to this issue). On the other hand, it is hard to reproduce this (and I have never got such error locally so far), and if code will changed I will never know if issue still remains or not.

Actually, right now I am working on #56 and it will already change some part of the code (related to `_requests` field updating), so this might already be modified.

Aaronontheweb:
@IgorFedchenko nah, I'd just keep it in mind for now and see if it happens again. Might be an issue you end up fixing as a result of other changes down the line.

IgorFedchenko:
Was still reproduced here: https://github.com/akkadotnet/Akka.Streams.Kafka/pull/79/checks?check_run_id=257039068

IgorFedchenko:
And once more here: https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=31981&view=ms.vss-test-web.build-test-results-tab

dimitar-j-shterev:
[...]

PR #79, opened 2019-10-11
Committable partitioned source

IgorFedchenko:
Next part of work on issue #36 , implements `CommittablePartitionedSource`.

Basically the same as `PlainPartitionedSource`, but emits `CommittableMessage` instead of just consumed records to downstream.


IgorFedchenko:
This PR is based on #65 , so it would be better to merge that one first.

Aaronontheweb:
@IgorFedchenko need a rebase here unfortunately

IgorFedchenko:
Again failed due to issue https://github.com/akkadotnet/Akka.Streams.Kafka/issues/57 - will just make it  perform new check for now.

PR #84, opened 2019-10-29
TransactionalSource

IgorFedchenko:
As next part of work for issue #36 , this PR is going to implement `TransactionalSource`, `TransactionalSink` and `TransactionalFlow`.

First of them is already implemented, but it can not be tested and used without one of the transactional producer stages. `TransactionalSink` as a simple wrapper over  `TransactionalFlow`, and both of them are used in tests for `TransactionalSource`. So, all three are going to be implemented in this PR to make CI checks useful.

Update: Only  `TransactionalSource` will be implemented here, and producer related code with tests will be in next PR.


IgorFedchenko:
Ah, I see now that producers implementation depends on #83 PR which is on review now. But currently implemented `TransactionalSource` does not depend on it, so let me make this PR ready for review without tests - it will be tested in another PR, which will make use if this one and #83.

Issue #85, opened 2019-10-30
Fix TransactionalProducer implementation once Confluent driver support this


IgorFedchenko:
There are couple of `TODO` labels in `TransactionalProducerStage.cs` file (in `TransactionalProducerStageLogic` class implementation). This is due to lack of transactional operations support in Confluent driver's producer implementation. 

This also requires some `librdkafka` update. The issue to track is https://github.com/confluentinc/confluent-kafka-dotnet/issues/1091.

Once this will be ready, implementation should work (there is a test  at `TransactionalIntegrationTests.cs` that is skipped right now).

IgorFedchenko:
Looks like they have published update which includes transaction support: https://www.nuget.org/packages/Confluent.Kafka/1.4.0

So now we can implement missing transactional stages.

chris-sung:
ah, this is a long waiting feature that I'm waiting to be implemented

IgorFedchenko:
@chris-sung Yeah, the problem is that `_producer.InitTransactions` call to Confluent driver is blocking for some reason, and I can't make my PR passing tests. And I ran out of time, not able to fix it now... If you will have time and willing to check it out or fork my branch and play with this, you are wellcome :)

Otherwise it may take a long time to have this working...

rafalpiotrowski:
Hi, Is this resolved now? or still pending?

IgorFedchenko:
Hey, nope, this is not resolved - there is a PR (https://github.com/akkadotnet/Akka.Streams.Kafka/pull/196) which was never finished (see comments above). Probably it will work now with latest driver versions, not sure.

PR #86, opened 2019-10-30
TransactionalFlow and TransactionalSink support

IgorFedchenko:
As a part of issue #36, transactional producer methods are implemented.

They should work in pair with `KafkaConsumer.TransactionalSource` for exactly once delivery guarantee.

Unfortunately, due to Confluent's driver [issue](https://github.com/confluentinc/confluent-kafka-dotnet/issues/1091) there are missing few important lines in implementation (they are marked with `TODO` labels). Without them, implementation does not work correctly and tests fail.

With that said, the implementation by itself is finished, and once the transactional support in Confluent driver will be finished we can quickly adopt it. I created issue for that: https://github.com/akkadotnet/Akka.Streams.Kafka/issues/85

For now, I am skipping failing test to let CI checks pass.

P.S. This PR is based on #83 and #84 PRs, so better to merge them first.


src/Akka.Streams.Kafka/Stages/Producers/TransactionalProducerStage.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            var groupId = batch.GroupId;
+            Log.Debug("Committing transaction for consumer group '{0}' with offsets: {1}", groupId, batch.Offsets);
+            var offsetMap = batch.OffsetMap();
+            
+            // TODO: Add producer work with transactions
+            /* scala code:
+             producer.sendOffsetsToTransaction(offsetMap.asJava, group)
+             producer.commitTransaction()
+             */
```
IgorFedchenko: Here is an example of missing lines, that should be replaced to close #85 . Few more places like this in next methods.

Aaronontheweb:
@IgorFedchenko more conflicts here

IgorFedchenko:
Oh yea, lots of them... Resolved now.

PR #100, opened 2020-01-31
PlainPartitionedSource_should_work racy test fix

IgorFedchenko:
After update to Confluent.Kafka 1.3.0 startup time for consumer increased to ~6 seconds for some reason. Maybe we shell check our docker image version, if it is outdated for this consumer version...

Anyway, one of the tests was very dependent on timings - so made it more robust.


Aaronontheweb:
> Maybe we shell check our docker image version, if it is outdated for this consumer version... 

Not a bad idea.

PR #112, opened 2020-03-10
Fixes for batch committer flow and commit refreshing

hyperlobic:
First of all, thanks for putting this project out there! I just started exploring it recently and discovered some unexpected behavior when trying to use Committer.Flow and Committer.Sink for batch commits with CommittableSource. I tracked it down to a couple issues.

The first thing was a missing assignment in CommittableOffsetBatch, resulting in the initial batch not getting created correctly.

After fixing that I noticed offsets would continually reset back to the initial batch. I narrowed this down to the offset management in the commit refreshing code. I switched some sets to dictionaries (which matches the Scala implementation slightly more closely) to resolve the unbounded growth of offsets.

Additionally, it appears the default for commit refreshing should be disabled, but instead the refresh interval ended up be zero, so refreshes were occurring on every poll iteration. I updated that, added the missing config setting to reference.conf, and changed the "disabled" value to match what HOCON is using for infinite.

Also included a integration test which is basically a variation on one of the tests for CommittableSource.

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/CommitterFlowIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/CommitterFlowIntegrationTests.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Akka.Configuration;
+using Akka.Streams.Dsl;
+using Akka.Streams.Kafka.Dsl;
+using Akka.Streams.Kafka.Messages;
+using Akka.Streams.Kafka.Settings;
+using Akka.Streams.TestKit;
+using Akka.Streams.Kafka.Helpers;
+using Confluent.Kafka;
+using Xunit;
+using Xunit.Abstractions;
+using Config = Hocon.Config;
+
+namespace Akka.Streams.Kafka.Tests.Integration
+{
+    public class CommitterFlowIntegrationTests : KafkaIntegrationTests
+    {
+        public CommitterFlowIntegrationTests(ITestOutputHelper output, KafkaFixture fixture) 
+            : base(nameof(CommitterFlowIntegrationTests), output, fixture)
+        {
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(5)]
+        public async Task CommitterFlow_commits_offsets_from_CommittableSource(int batchSize)
+        {
+            var topic1 = CreateTopic(1);
+            var topicPartition1 = new TopicPartition(topic1, 0);
+            var group1 = CreateGroup(1);
+
+            await GivenInitializedTopic(topicPartition1);
+
+            await Source
+                .From(Enumerable.Range(1, 100))
+                .Select(elem => new ProducerRecord<Null, string>(topicPartition1, elem.ToString()))
+                .RunWith(KafkaProducer.PlainSink(ProducerSettings), Materializer);
+
+            var consumerSettings = CreateConsumerSettings<string>(group1);
+            var committedElements = new ConcurrentQueue<string>();
+            var committerSettings = CommitterSettings.WithMaxBatch(batchSize);
+            
+            var (task, probe1) = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.Assignment(topicPartition1))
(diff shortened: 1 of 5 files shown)


src/Akka.Streams.Kafka.Tests/Integration/CommitterFlowIntegrationTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            var topic1 = CreateTopic(1);
+            var topicPartition1 = new TopicPartition(topic1, 0);
+            var group1 = CreateGroup(1);
+
+            await GivenInitializedTopic(topicPartition1);
+
+            await Source
+                .From(Enumerable.Range(1, 100))
+                .Select(elem => new ProducerRecord<Null, string>(topicPartition1, elem.ToString()))
+                .RunWith(KafkaProducer.PlainSink(ProducerSettings), Materializer);
```
IgorFedchenko: There is a helpful `ProduceStrings` method in base class to make this in one line - you may want to use it here

src/Akka.Streams.Kafka.Tests/Integration/CommitterFlowIntegrationTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            foreach (var _ in Enumerable.Range(1, 25 / batchSize))
+            {
+                probe1.ExpectNext(Done.Instance, TimeSpan.FromSeconds(10));
+            }
+                
+            probe1.Cancel();
+
+            AwaitCondition(() => task.IsShutdown.IsCompletedSuccessfully);
+
+            var probe2 = KafkaConsumer.PlainSource(consumerSettings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
```
IgorFedchenko: `new TopicPartition(topic1, 0)` is basically `topicPartition1 ` defined above, right?

src/Akka.Streams.Kafka/Helpers/CommitRefreshing.cs
```diff
@@ -25,7 +27,7 @@ internal static class CommitRefreshing
     {
         public static ICommitRefreshing<K, V> Create<K, V>(TimeSpan commitRefreshInterval)
         {
-            if (commitRefreshInterval == TimeSpan.MaxValue)
+            if (commitRefreshInterval == Timeout.InfiniteTimeSpan)
```
IgorFedchenko: Huh, didn't know that HOCON parser uses this... Actually, did not check that out. Thanks!

src/Akka.Streams.Kafka/Helpers/CommitRefreshing.cs
```diff
@@ -46,24 +48,26 @@ public class NoOp<K, V> : ICommitRefreshing<K, V>
         public class Impl<K, V> : ICommitRefreshing<K, V>
         {
             private readonly TimeSpan _commitRefreshInterval;
-            private IImmutableSet<TopicPartitionOffset> _requestedOffsets = ImmutableHashSet<TopicPartitionOffset>.Empty;
-            private IImmutableSet<TopicPartitionOffset> _committedOffsets = ImmutableHashSet<TopicPartitionOffset>.Empty;
+            private IImmutableDictionary<TopicPartition, Offset> _requestedOffsets = ImmutableDictionary<TopicPartition, Offset>.Empty;
+            private IImmutableDictionary<TopicPartition, Offset> _committedOffsets = ImmutableDictionary<TopicPartition, Offset>.Empty;
```
IgorFedchenko: Yup, scala uses dictionaries in few places where I was using just sets with `TopicPartitionOffset` objects. Thanks for updating here, since looks like code becomes simpler in this case.

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
```diff
@@ -143,7 +142,7 @@ private ICommittableOffsetBatch UpdateWithOffset(ICommittableOffset committableO
             }
             else
             {
-                Committers.SetItem(partitionOffset.GroupId, committer);
+                newCommitters = Committers.SetItem(partitionOffset.GroupId, committer);
```
IgorFedchenko: Woops - great catch!

PR #167, opened 2020-12-09
Make kafka work with event hub

Arkatufus:
- Strict consumer creation, don't supress exception when Kafka failed to create a consumer.
- Some quality of life code shortcuts
- Add Kafka to Event Hub connection sample


IgorFedchenko (approved):
LGTM

Issue #184, opened 2021-03-04
Improve Kafka error handling


Arkatufus:
Need to improve how the producer code handles Kafka Error report by wrapping `IProducer<K, V>.Produce` with try...catch and only fail the stage based on an exception thrown from there. 
Error object coming from the delivery report should not cause a failed stage.

Aaronontheweb:
What happens in these scenarios, per Kafka documentation - vs. what happens in our library?

* what happens when you reach the end of a partition?
* when happens when offsets get TTLed by Kafka's retention system in-flight?
* what happens when a connection goes idle and gets reaped by the broker?

Arkatufus:
Documentation burried in the kafka client [documentation](https://docs.confluent.io/5.5.1/clients/librdkafka/rdkafka_8h.html):

### Transactional producer API

The transactional producer  operates on top of the idempotent producer, and provides full  exactly-once semantics (EOS) for Apache Kafka when used with the  transaction aware consumer (`isolation.level=read_committed`). 

A producer instance is configured for transactions by setting the `transactional.id` to an identifier unique for the application. This id will be used to  fence stale transactions from previous instances of the application,  typically following an outage or crash. 

After creating the transactional producer instance using `rd_kafka_new()` the transactional state must be initialized by calling `rd_kafka_init_transactions()`.  This is a blocking call that will acquire a runtime producer id from  the transaction coordinator broker as well as abort any stale  transactions and fence any still running producer instances with the  same `transactional.id`.

Once transactions are initialized the application may begin a new transaction by calling `rd_kafka_begin_transaction()`. A producer instance may only have one single on-going transaction. 

Any messages produced after the transaction has been started will  belong to the ongoing transaction and will be committed or aborted  atomically. It is not permitted to produce messages outside a  transaction boundary, e.g., before `rd_kafka_begin_transaction()` or after `rd_kafka_commit_transaction()`, `rd_kafka_abort_transaction()`, or after the current transaction has failed. 

If consumed messages are used as input to the transaction, the consumer instance must be configured with `enable.auto.commit` set to `false`.  To commit the consumed offsets along with the transaction pass the list  of consumed partitions and the last offset processed + 1 to `rd_kafka_send_offsets_to_transaction()` prior to committing the transaction. This allows an aborted transaction to be restarted using the previously committed offsets. 

[...]

Issue #33, opened 2019-08-12
Provide defaults for consumer supervision strategy on Restart


IgorFedchenko:
Right now we have error handling like this:
https://github.com/IgorFedchenko/Akka.Streams.Kafka/blob/02d0f046a76482a444481a681175917f526c5b89/src/Akka.Streams.Kafka/Stages/CommittableConsumerStage.cs#L165-L183

So, when decision is to restart stage, we need to something else then just resume.

@Aaronontheweb I guess we need to dispose current consumer and create another one here. 
I am still little bit new with Akka.Streams - so is there some handy way to stop stage and start again without delivering failure to the user? This will do exactly what we need - will do consumer.Dispose() in PostStop call and then recreate it.

Aaronontheweb:
There are some built-in stages that can do this - how does the original Alpakka-Kafka project do this?

IgorFedchenko:
Got your point, will check this out when return to this issue.

PR #185, opened 2021-03-09
Use Decider for choosing producer background error handling 

IgorFedchenko:
Related to #184 and #33


src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -233,7 +232,18 @@ private void HandleProduceError(IProducer<K, V> producer, Error error)
...
                 var exception = new KafkaException(error);
-                FailStage(exception);
+                switch (_decider(exception))
+                {
+                    case Directive.Stop:
+                        if (_stage.CloseProducerOnStop)
+                        {
+                            producer.Dispose();
+                        }
+                        _failStageCallback(exception);
```
IgorFedchenko, line 242: Need to have `Directive.Restart` support too - but it is better then what we had anyway.

PR #190, opened 2021-03-15
Add ConnectionChecker feature

Arkatufus:
- Add ConnectionChecker
- Add the ability to declare kafka settings directly inside of HOCON (declaring key and value serializer/deserializer is not implemented yet)

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/KafkaIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/KafkaIntegrationTests.cs
@@ -150,10 +150,8 @@ namespace Akka.Streams.Kafka.Tests
 
         private static Config Default()
         {
-            var defaultSettings =
-                ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf");
-            
-            var config = ConfigurationFactory.ParseString("akka.loglevel = DEBUG");
+            //var config = ConfigurationFactory.ParseString("akka.loglevel = DEBUG");
+            var config = ConfigurationFactory.ParseString("akka{}");
 
             if (TestsConfiguration.UseFileLogging)
             {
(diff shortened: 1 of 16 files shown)


src/Akka.Streams.Kafka.Tests/KafkaIntegrationTests.cs
```diff
@@ -150,18 +150,16 @@ protected async Task GivenInitializedTopic(TopicPartition topicPartition)
 
         private static Config Default()
         {
-            var defaultSettings =
-                ConfigurationFactory.FromResource<ConsumerSettings<object, object>>("Akka.Streams.Kafka.reference.conf");
-            
-            var config = ConfigurationFactory.ParseString("akka.loglevel = DEBUG");
+            //var config = ConfigurationFactory.ParseString("akka.loglevel = DEBUG");
+            var config = ConfigurationFactory.ParseString("akka{}");
```
IgorFedchenko, line 154: Why not to keep Debug level here? I guess this is helpful for debugging by default.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -168,15 +183,24 @@ protected override bool Receive(object message)
...
                 
-                case KafkaConsumerActorMetadata.Internal.Stop stop:
+                case KafkaConsumerActorMetadata.Internal.Stop _:
                     _log.Debug($"Received Stop from {Sender}, stopping");
                     Context.Stop(Self);
                     return true;
                 
+                case KafkaConnectionFailed kcf:
+                    ProcessError(kcf);
+                    Self.Tell(KafkaConsumerActorMetadata.Internal.Stop.Instance);
```
IgorFedchenko, line 193: So if there is a temporary disconnect from broker, there is no other way of handling that except stopping consuming? Right now this depends on supervision strategy, you can just try to resume. ProcessError method already sending failure notification to parent (source stage), and parent may stop this actor if `Dicider` is configured to do that.

IgorFedchenko (commented):
LGTM, but left one question about stopping actor instaed of letting parent stage to dicide what to do using supervision strategy.

IgorFedchenko:
Added an issue to update documentation https://github.com/akkadotnet/Akka.Streams.Kafka/issues/191

PR #193, opened 2021-03-23
Use Consumer.Close in code defensively

Arkatufus:


src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -259,24 +259,36 @@ private void ApplySettings(ConsumerSettings<K, V> updatedSettings)
 
         protected override void PostStop()
         {
-            if (_settings.ConnectionCheckerSettings.Enabled)
+            base.PostStop();
```
IgorFedchenko, line 262: So in alpakka base `PostStop` is executed before the rest of the handling?

IgorFedchenko (commented):
Left some questions about changes in `KafkaConsumerActor.PostStop` logic, which are not related to using `Close` method. 

As you mentioned before - the more patches we apply, the harder to keep stuff in sync with scala version. Does alpakka has this `PostStop` method implemented like this? Because to me, there is no reason to update this method for now, unless it does not match alpakka implementation (at the moment of porting, it was matching I think).

IgorFedchenko (approved):
Let's put it like this @Arkatufus : if you think that this change gives more benefits then takes (I mean the difference with alpakka code) - merge it, since as is it looks good to me, and passes all the tests.

Aaronontheweb:
@IgorFedchenko @Arkatufus have conflicts here

PR #196, opened 2021-03-25
Make TransactionalProducer implementation working

IgorFedchenko:
Closes https://github.com/akkadotnet/Akka.Streams.Kafka/issues/85


IgorFedchenko:
Looks like there is some issue with `TransactionalProducer.Sink` here - `InitTransactions` call is hanging, which breaks test.

PR #222, opened 2021-08-31
Update RELEASE_NOTES.md for 1.1.4 release

Arkatufus:


Arkatufus:
These racy tests are really bugging me now, it might be caused by a real underlying problem

PR #225, opened 2021-09-07
Fix failing Kafka seek causing plain partitioned source to fail

Arkatufus:


Arkatufus:
Added a native client only test that simulates the same behaviour with our test.
The final verdict is yes, doing a seek on a topic partition is really slow, it takes about 500ms for a seek to complete.
Final number for the benchmark is approximately 9 seconds, our test number is approximately 10 seconds, the benchmark doesn't have to deal with threading, stream, etc.

PR #241, opened 2021-11-05
Significantly improved efficiency of `Poll` loop

Aaronontheweb:

PR #246, opened 2021-11-09
Fix SupervisionStrategy decider isn't being called properly

### LINKED ISSUE #240: SupervisionStrategy does not restart Akka.Streams.Kafka Sources when Downstream Fails
FWIW, unrelated to this, but came up during my investigation: ` .WithAttributes(ActorAttributes.CreateSupervisionStrategy(decider))` does not help restart failed Kafka `Source` stages when a downstream fails, which seems off to me. Going to file a separate bug for it.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/Akka.Streams.Kafka/issues/232#issuecomment-962112473_

Arkatufus:
Closes #240


src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -118,45 +118,35 @@ protected virtual void MessageHandling((IActorRef, object) args)
...
-                    switch (_decider(failure.Cause))
-                    {
-                        case Directive.Stop:
-                            // Throw
-                            FailStage(exception);
-                            break;
-                        case Directive.Resume:
-                            // keep going
-                            break;
-                        case Directive.Restart:
```
Aaronontheweb, line 135: LOL..... well that would do it.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -118,45 +118,35 @@ protected virtual void MessageHandling((IActorRef, object) args)
...
-                            // TODO: Need to do something here: https://github.com/akkadotnet/Akka.Streams.Kafka/issues/33
-                            break;
-                    }
+                    FailStage(failure.Cause);
                     break;
                 
-                case Terminated terminated:
+                case Terminated terminated when ReferenceEquals(terminated.ActorRef, ConsumerActor):
```
Aaronontheweb, line 127: Why is the `ReferenceEquals` needed here?

Aaronontheweb: What other actors are being `Context.Watch`'d by this actor?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -118,45 +118,35 @@ protected virtual void MessageHandling((IActorRef, object) args)
...
                     FailStage(new ConsumerFailed());
                     break;
             }
         }
 
         private void Pump()
         {
-            if (IsAvailable(_shape.Outlet))
+            while (true)
```
Aaronontheweb, line 135: This will conflict with https://github.com/akkadotnet/Akka.Streams.Kafka/pull/241

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActorMetadata.cs
```diff
@@ -13,7 +13,7 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Actors
     /// Generally this should not be used from outside of the library.
     /// </summary>
     [InternalApi]
-    public class KafkaConsumerActorMetadata
+    public static class KafkaConsumerActorMetadata
```
Aaronontheweb, line 16: LGTM

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -138,7 +131,7 @@ public override void PreStart()
         {
             base.PreStart();
 
-            Producer = _stage.ProducerProvider(HandleProduceError);
+            Producer = _stage.ProducerProvider(null);
```
Aaronontheweb, line 134: Does `HandleProduceError` not get called anymore?

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -225,28 +220,58 @@ private void CheckForCompletion()
                 }
             }
         }
-        
-        private void HandleProduceError(IProducer<K, V> producer, Error error)
+
+        private void CloseAndFailStage(Exception ex)
         {
-            Log.Error(error.Reason);
```
Aaronontheweb, line 231: Does this get logged somewhere else?

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -225,28 +220,58 @@ private void CheckForCompletion()
...
+            CloseProducerImmediately();
+            FailStage(ex);
+        }
+
+        private void CloseProducerImmediately()
```
Aaronontheweb, line 230: Maybe change the method name if this method might not actually close the producer.

Aaronontheweb (changes requested):
Where does `SubSource` derive from these classes?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: It has to do with clean shutdown ordering

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: i'll revert it then

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
Arkatufus: Nope, we don't care about it. everything is handled inside the new message received callback. It is the same thing, the error handler is called before the message is passed to the message received callback.

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
Arkatufus: Yes, in the message received callback

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Aaronontheweb: What other actors could send a `Terminate` message though?

Aaronontheweb:
Where does `SubSource` derive from these classes @Arkatufus ?

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: This method is not being used in this class, it is being passed as the receive method of the StageActor

Arkatufus:
So far it shares the same KafkaConsumerActor implementation, but with different parameters passed in.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: If you look in the PreStart(), we explicitly told the StageActor instance to watch the KafkaConsumerActor instance.

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
Arkatufus: Well, it will dispose the producer immediately if it owns the producer. If it is an external producer, we skip the dispose. Difference with the other close method is the other method will attempt to flush the producer first before disposing it.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            probe.ExpectNoMsg(TimeSpan.FromSeconds(5));
+            probe.Cancel();
+            
+            // restart dead stream
+            probe = KafkaConsumer.CommittableSource(consumerSettings, Subscriptions.AssignmentWithOffset(new TopicPartitionOffset(topicPartition, Offset.Unset)))
+                .Select(t =>
+                {
+                    counter++;
+                    if (counter % 7 == 0)
+                        throw new Exception("BOOM!");
```
Aaronontheweb, line 176: Need to add a comment explaining why this doesn't throw again when we re-create the stream.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                callCount++;
+                return Directive.Resume;
+            }
+
+            var committerSettings = CommitterSettings.Create(Sys);
+            var consumerSettings = CreateConsumerSettings<string>(group);
+            var counter = 0;
+
+            await Source.From(Enumerable.Range(1, 11))
```
Aaronontheweb, line 217: Add a comment here explaining that this is part of the "arrange" stage of the test to help populate data we will read during our "act" stage

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            }
+
+            var topic1 = CreateTopic(1);
+            var group1 = CreateGroup(1);
+            
+            var producerSettings = ProducerSettings<Null, string>
+                .Create(Sys, null, new FailingSerializer())
+                .WithBootstrapServers(Fixture.KafkaServer);
+            
+            var sourceTask = Source
```
Aaronontheweb, line 280: Add comments here to explain how failures are injected

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            var topic1 = CreateTopic(1);
+            var group1 = CreateGroup(1);
+
+            var sourceTask = ProduceStrings(new TopicPartition(topic1, 0), Enumerable.Range(1, elementsCount), ProducerSettings);
+
+            await GuardWithTimeoutAsync(sourceTask, TimeSpan.FromSeconds(3));
+            
+            var settings = CreateConsumerSettings<Null, string>(group1).WithAutoCreateTopicsEnabled(false);
+
+            var probe = KafkaConsumer
```
Aaronontheweb, line 355: Need to explain that we're subscribing to a non-existent partition, which creates the resulting error from the Confluent driver

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+                .PlainSource(settings, Subscriptions.Assignment(new TopicPartition(topic1, 5)))
+                .Select(c => c.Value)
+                .RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.Request(elementsCount);
+            var error = probe.ExpectEvent(TimeSpan.FromSeconds(5));
+            error.Should().BeOfType<TestSubscriber.OnError>();
+            var exception = ((TestSubscriber.OnError)error).Cause;
+            exception.Should().BeOfType<KafkaException>();
+            ((KafkaException) exception).Error.Code.Should().Be(ErrorCode.Local_UnknownPartition);
```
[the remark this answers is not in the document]
Aaronontheweb: We can address this in a separate PR

src/Akka.Streams.Kafka.Tests/Integration/PlainPartitionedSourceIntegrationTests.cs
```diff
@@ -178,8 +178,13 @@ public async Task PlainPartitionedSource_should_be_signalled_about_serialization
             substream.Request(1);
             
             await ProduceStrings(topic, new int[] { 0 }, ProducerSettings); // Produce "0" string
-            
-            Within(TimeSpan.FromSeconds(10), () => substream.ExpectError().Should().BeOfType<SerializationException>());
+
+            Within(TimeSpan.FromSeconds(10), () =>
+            {
+                var err = substream.ExpectError();
+                err.Should().BeOfType<ConsumeException>();
```
Aaronontheweb, line 185: Changed to reflect how we handle downstream errors.

src/Akka.Streams.Kafka.Tests/Integration/PlainSinkIntegrationTests.cs
```diff
@@ -88,5 +91,76 @@ public async Task PlainSink_should_fail_stage_if_broker_unavailable()
             probe.ExpectSubscription();
             probe.OnError(new KafkaException(ErrorCode.Local_Transport));
         }
+        
+        [Fact]
+        public async Task PlainSink_should_resume_on_deserialization_errors()
```
Aaronontheweb, line 96: Should make it clear that we're testing a custom decider here, not the built-in one.

src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
```diff
@@ -148,16 +148,23 @@ public async Task PlainSource_should_stop_on_deserialization_errors()
...
         [Fact]
         public async Task PlainSource_should_resume_on_deserialization_errors()
         {
-            Directive Decider(Exception cause) => cause is SerializationException
-                ? Directive.Resume
-                : Directive.Stop;
+            var callCount = 0;
+            Directive Decider(Exception cause)
+            {
+                if(cause is ConsumeException ex && ex.Error.IsSerializationError())
```
Aaronontheweb, line 161: So when there's a deserialization error on consume, do we still get a `ConsumeResult` that we could pass downstream?

Aaronontheweb: That way, if someone wanted to commit / discard the bad message and advance the consumer - how would they go about doing that? This is something we'd need to have an example of in our code / documentation.

Aaronontheweb: Might want to mention schema registry as a general preventative solution to this here.

src/Akka.Streams.Kafka/Internal/ConnectionChecker.cs
```diff
@@ -84,6 +84,7 @@ protected override void PreStart()
         {
             base.PreStart();
             StartTimer();
+            Become(Regular());
```
Aaronontheweb, line 87: Is this a bugfix?

src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
```diff
@@ -67,7 +68,9 @@ public sealed class ConsumerSettings<TKey, TValue>
                 dispatcherId: config.GetString("use-dispatcher", "akka.kafka.default-dispatcher"),
                 autoCreateTopicsEnabled: config.GetBoolean("allow.auto.create.topics", true),
                 properties: properties,
-                connectionCheckerSettings: ConnectionCheckerSettings.Create(config.GetConfig(ConnectionCheckerSettings.ConfigPath)));
+                connectionCheckerSettings: ConnectionCheckerSettings.Create(config.GetConfig(ConnectionCheckerSettings.ConfigPath)),
+                closeTimeout: config.GetTimeSpan("stop-timeout"),
+                consumerFactory: null );
```
Aaronontheweb, line 73: What is this for?

src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
```diff
@@ -360,4 +415,16 @@ public sealed class ConsumerSettings<TKey, TValue>
                 .Build();
         }
     }
+
+    internal sealed class RebalanceListener<TKey, TValue>
```
Aaronontheweb, line 419: What's this for? I thought we already had a rebalance listener implementation.

src/Akka.Streams.Kafka/Settings/KafkaExtensions.cs
```diff
@@ -9,41 +9,11 @@ public static class KafkaExtensions
...
-                case ErrorCode.NotLeaderForPartition:
-                case ErrorCode.RequestTimedOut:
-                case ErrorCode.GroupLoadInProress:
-                case ErrorCode.GroupCoordinatorNotAvailable:
-                case ErrorCode.NotCoordinatorForGroup:
-                case ErrorCode.NotEnoughReplicas:
-                case ErrorCode.NotEnoughReplicasAfterAppend:
-                    return true;
-            }
+        public static bool IsSerializationError(this Error error)
```
Aaronontheweb, line 12: LGTM

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -119,28 +119,34 @@ protected virtual void MessageHandling((IActorRef, object) args)
...
+                    // to prevent any offset/commit runaway. We will need to restart it.
+                    SourceActor.Unwatch(ConsumerActor);
+                    ConsumerActor = CreateConsumerActor();
+                    SourceActor.Watch(ConsumerActor);
+                    ConfigureSubscription();
                     break;
                 
                 case Terminated terminated:
-                    FailStage(new ConsumerFailed());
+                    Log.Info($"Consumer actor terminated: {terminated.ActorRef.Path}");
```
Aaronontheweb, line 149: So we don't terminate anymore when the consumer actor dies?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -321,13 +321,21 @@ private void ApplySettings(ConsumerSettings<K, V> updatedSettings)
...
-                    consumeErrorHandler: (c, e) => ProcessError(new KafkaException(e)),
-                    partitionAssignedHandler: (c, tp) => Self.Tell(new PartitionAssigned(tp.ToImmutableHashSet())),
-                    partitionRevokedHandler: (c, tp) => Self.Tell(new PartitionRevoked(tp.ToImmutableHashSet())),
+                    consumeErrorHandler: (c, e) =>
+                    {
+                        var exception = new KafkaException(e);
+                        ProcessError(exception);
+                        _pollCancellation?.Cancel();
+                        _log.Error(exception, "Exception when polling from consumer, stopping actor: {0}", exception.ToString());
+                        Context.Stop(Self);
```
Aaronontheweb, line 332: It would have to be a pretty severe error from the consumer to trigger this, right? Something like our idle connections being reaped would not cause this - correct?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -625,7 +613,6 @@ private void ProcessError(Exception error)
             foreach (var actor in involvedStageActors)
             {
                 actor.Tell(new Status.Failure(error));
-                _requests = _requests.Remove(actor);
```
Aaronontheweb, line 628: Do we not need to track requests anymore?

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -64,9 +61,6 @@ public DefaultProducerStageLogic(IProducerStage<K, V, P, TIn, TOut> stage, Attri
 
             var supervisionStrategy = attributes.GetAttribute<ActorAttributes.SupervisionStrategy>(null);
             _decider = supervisionStrategy != null ? supervisionStrategy.Decider : Deciders.StoppingDecider;
-            
-            _checkForCompletionCallback = GetAsyncCallback(CheckForCompletion);
```
Aaronontheweb, line 68: Can you, generally, comment on why you made these changes to this file?

Aaronontheweb (changes requested):
Looking better but need to work on some maintainability issues and add comments to help explain the non-obvious to other contributors and OSS users

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: In the event of a serialization error, the KafkaConsumerActor does not get killed. Removing all requests will make the consumer actor and the downstream stage to hang if the user decides to ignore the serialization error.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: Incorrect, all errors are being processed, the actor kills itself after sending the failure message to the stage to be processed.
It is the job of the stage to decide if it needs to stop itself or restart the dead actor.

We keep on polling for messages even when there are no requests from downstream, this is done safely by pausing all subscribed topics before we do the poll calls, this ensures that only system messages are being processed by these blank polls.

The consequence is that these errors, while it usually happened during consume in normal kafka applications, will appear as if it happened "out of band", and some errors like the "all brokers are down" errors would not be propagated downstream, since there will be no requests pending, which is quite safe.

But we do need to make a custom decider on the stage itself that blocks these kind of nonsense "errors" and not have the stream crash and burn while there are no messages on the queue.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: No, we don't terminate the stage on consumer actor death, that will be handled by the decider.
The decider depends solely on the failure messages being passed from the actor and decides if it needs to re-create the dead actor or to fail the stage.

src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
Arkatufus: This is an internal class used only in tests. Its job is to pass the internal callback handlers set up inside the consumer actor to the mock consumer to emulate a real callbacks during runtime.

src/Akka.Streams.Kafka/Internal/ConnectionChecker.cs
Arkatufus: Yes, its a bug fix, discovered by the newly added ConnectionChecker unit tests

src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
Arkatufus: We can't pass the failed de/serialized message downstream because it is of a different type compared to the stream type.
When a message failed to be de/serialized, it is passed inside the ConsumeException as a ConsumeResult<byte[], byte[]>. The only way to examine these messages is by catching it in the Supervision decider or to create the stream without any key/value de/serializer by making it a stream of <byte[], byte[]> instead.

Arkatufus: To ignore/discard failed de/serialized messages, the user can simply use a decider that always resumes the stream when a de/serialization error occured.

src/Akka.Streams.Kafka.Tests/Integration/CommitterFlowIntegrationTests.cs
```diff
@@ -24,8 +24,7 @@ public CommitterFlowIntegrationTests(ITestOutputHelper output, KafkaFixture fixt
         {
         }
 
-        // This test is very very specific and will only work with batchSize values of either 1 or 5.
-        [Theory]
+        [Theory(Skip = "This test looks fishy, will need to investigate -- Gregorius")]
```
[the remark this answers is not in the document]
Arkatufus: Already filed as an issue at https://github.com/akkadotnet/Akka.Streams.Kafka/issues/251

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+                .PlainSource(settings, Subscriptions.Assignment(new TopicPartition(topic1, 0)))
+                .WithAttributes(ActorAttributes.CreateSupervisionStrategy(Deciders.StoppingDecider))
+                .Select(c => c.Value)
+                .RunWith(this.SinkProbe<int>(), Materializer);
+
+            var error = probe.Request(elementsCount).ExpectEvent(TimeSpan.FromSeconds(5));
+            error.Should().BeOfType<TestSubscriber.OnError>();
+            var exception = ((TestSubscriber.OnError)error).Cause;
+            exception.Should().BeOfType<ConsumeException>();
+            ((ConsumeException) exception).Error.IsSerializationError().Should().BeTrue();
```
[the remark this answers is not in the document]
Arkatufus: They do terminate the stream, but they left the KafkaConsumerActor running, which is kind of weird, because they did not implement any supervision decider on their consumer/source stage. Their default is to crash the stage if it happens and let the stages downstream to pick up the error and handle it.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+
+            probe.Request(10);
+            for (var i = 0; i < 9; i++)
+            {
+                var message = probe.ExpectNext(TimeSpan.FromSeconds(10)); 
+                Log.Info(message);
+            }
+            probe.Cancel();
+            
+            callCount.Should().Be(2);
```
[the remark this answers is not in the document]
Arkatufus: Changed the failing code so it only happens once, it was 2 because we're sending 11 data and exceptions happends when mod 5 returns 0

src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
Aaronontheweb: > To ignore/discard failed de/serialized messages, the user can simply use a decider that always resumes the stream when a de/serialization error occured.

Still need to commit the offset somewhere - so that will need to be documented for stages that don't automatically commit offsets i.e. `CommitableSource`

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Aaronontheweb: And the read buffer is stored on the actor or inside the stage? I had thought it was the latter

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: > But we do need to make a custom decider on the stage itself that blocks these kind of nonsense "errors" and not have the stream crash and burn while there are no messages on the queue.

Agree

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: The `actor` in this case was the `StageActor` for the graph stage though, was it not?

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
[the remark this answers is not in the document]
Aaronontheweb: Can one `KafkaConsumerActor` get used by multiple stages in the event of a `SubSource`?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: No, its the KafkaConsumerActor, the StageActor are only used as a middle man to pass messages back from the KafkaConsumerActor to the stage using its Receive method. StageActor will never be killed in an exception.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: Buffer is held inside the stage, that gets purged when the stage is told to resume or restart.
The Kafka consumer lives inside the actor, that held the offset counter for the current session, that needs to be killed to make sure that we're restarting the offset on a resume or a restart.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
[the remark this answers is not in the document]
Arkatufus: Yes, KafkaConsumerActor is shared with all of the SubSource, they watch the actor and fails their stage if the actor died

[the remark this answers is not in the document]
Aaronontheweb: And the actor can only terminate itself is the `Consumer` throws a fatal exception? Or is it when the consumer throws  _any_ exception?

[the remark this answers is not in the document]
Aaronontheweb: In other words, can that 10/10 brokers down error cause this actor to die?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: Why would the `KafkaConsumerActor` track requests for itself?

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -171,39 +193,60 @@ public override void PostStop()
         
         protected virtual void PostSend(IEnvelope<K, V, P> msg) { }
 
-        private Action<DeliveryReport<K, V>> BuildSendCallback<TResult>(TaskCompletionSource<TResult> completion, Action<DeliveryReport<K, V>> onSuccess)
+        private void OnProduceFailure(Exception ex)
```
Arkatufus, line 196: All failures are routed here to be processed by the decider

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -171,39 +193,60 @@ public override void PostStop()
...
+                        }
+                    }
+                    TryPull(_stage.In);
+                    ConfirmAndCheckForCompletion();
+                    break;
+                }
+            }
+        }
+        
+        private Action<DeliveryReport<K, V>> BuildSendCallback<TResult>(
```
Arkatufus, line 224: Aggregated all callbacks into a single call. Before, each callback methods are decorated with the GetAsyncCallback inside this method. Now all of them are regular methods, we call the GetAsyncCallback decorator on this method instead. Less headache to maintain.

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -94,15 +99,34 @@ public DefaultProducerStageLogic(IProducerStage<K, V, P, TIn, TOut> stage, Attri
...
-                            });
-                            PostSend(msg);
-                            var resultTask = Task.WhenAll(tasks).ContinueWith(t => new MultiResult<K, V, P>(t.Result.ToImmutableHashSet(), multiMessage.PassThrough) as IResults<K, V, P>);
-                            Push(stage.Out, resultTask as Task<TOut>);
+                                    Producer.Produce(record, GetAsyncCallback(callback));
+                                    return result.Task;
+                                }
+                                catch (Exception exception)
+                                {
+                                    OnProduceFailure(exception);
```
Arkatufus, line 116: Enclose the Producer.Produce with a try..catch and route any exceptions to the OnProduceFailure method.

src/Akka.Streams.Kafka/Stages/Producers/DefaultProducerStage.cs
```diff
@@ -79,12 +73,23 @@ public DefaultProducerStageLogic(IProducerStage<K, V, P, TIn, TOut> stage, Attri
...
+                                        result.SetResult(new Result<K, V, P>(report, message));
+                                    }, 
+                                    onFailure: OnProduceFailure);
+                                Producer.Produce(message.Record, GetAsyncCallback(callback));
+                                PostSend(msg);
+                                Push(stage.Out, result.Task as Task<TOut>);
+                            }
+                            catch (Exception exception)
+                            {
+                                OnProduceFailure(exception);
```
Arkatufus, line 91: Enclose the Producer.Produce with a try..catch and route any exceptions to the OnProduceFailure method.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: They are used to track the requests made from the stages, only one for a normal stage, multiple for subsource stages.
When a stage is pulling, it sends a RequestsMessages message to the consumer actor which contains the topic and partition the stage is interested in. The consumer actor uses these to route messages it received from kafka to the interested stages.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+                    .Async()
+                    .GroupedWithin(1000, TimeSpan.FromSeconds(1))
+                    .Select(c => c.Count())
+                    .Log("MsgCount").AddAttributes(Attributes.CreateLogLevels(LogLevel.InfoLevel))
+                    .To(Sink.Ignore<int>()))
+                .Run(Sys.Materializer());
+
+            await Task.Delay(TimeSpan.FromSeconds(2));
+            await drainingControl.Shutdown();
+            callCount.Should().BeGreaterThan(0);
```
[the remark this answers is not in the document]
Arkatufus: Fixed

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
Arkatufus: Fixed

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            Directive Decider(Exception cause)
+            {
+                callCount++;
+                return Directive.Resume;
+            }
+
+            var consumerSettings = CreateConsumerSettings<string>(group);
+            var numbers = Source.From(new []{ 9,8,7,6,0,5,4,3,2,1 });
+            await numbers
+                .Via(Flow.Create<int>().Select(x => $"1/{x} is {1/x} w/ integer division"))
```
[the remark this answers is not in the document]
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/Integration/PlainSinkIntegrationTests.cs
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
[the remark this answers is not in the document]
Arkatufus: Yes, it will die. But the decider in the stage will restart it.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                var message = probe.ExpectNext();
+                Log.Info($"> [{i}]: {message}");
+            }
+            callCount.Should().Be(1);
+            probe.Cancel();
+        }
+        
+        [Fact]
+        public async Task SupervisionStrategy_Decider_on_PlainSource_should_work()
```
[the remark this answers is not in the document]
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
Arkatufus: Done

src/Akka.Streams.Kafka.Tests/Integration/PlainSourceIntegrationTests.cs
Arkatufus: Offsets are stored in the kafka consumer client, the de/serialization error is the only special case where the actor is not killed when it happened. Setting the decider to resume will automatically skip the failed message.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: That would mean that the `actor` in this case is the `StageActor` then right? Otherwise this code would make no sense.

Aaronontheweb: i.e. the `HashSet<T>` would max out at 1 item and there wouldn't be any purpose in supporting tracking requests from multiple stages.

Arkatufus: Oh, right, yes, `actor` in this context is the StageActor. `KafkaConsumerActor` is sending all of the messages to the stage `StageActor`, the stage registers a `Receive` on their `StageActor` to read all those messages

Aaronontheweb: Ok, got it confirmed in our chat - these are indeed requests being tracked to individual `StageActor`s, which makes sense.

Aaronontheweb (approved):
LGTM

Issue #228, opened 2021-09-08
CommitableSource consumer routed via FlexiFlow producer seemed to disconnect after awhile.


Arkatufus:
User reported via Gitter that their code gets disconnected after a period of time and eventually everything stops; the StreamSupervisor stops everything and it doesn't recover.

The logs are:
```
%6|1631044923.925|FAIL|my-app#consumer-8| [thrd:sasl_ssl://[REDACTED]:9092/7]: sasl_ssl://[REDACTED]:9092/7: Disconnected (after 2670559ms in state UP)

Sending failure to [akka://my-app/user/StreamSupervisor-0/$$d#539697147]. Error: Confluent.Kafka.KafkaException: sasl_ssl://[REDACTED]:9092/7: Disconnected (after 2670559ms in state UP)

"Received Stop from [akka://my-app/user/StreamSupervisor-0/$$d#539697147], stopping","@i":"db903a45","@l":"Debug","Message":"Received Stop from [akka://my-app/user/StreamSupervisor-0/$$d#539697147], stopping","Thread":"0007","LogSource":"akka://my-app/system/kafka-consumer-5","Timestamp":"2021-09-07T20:02:03.9997609Z","ActorPath":"akka://my-app/system/kafka-consumer-5","SourceContext":"Akka.Streams.Kafka.Stages.Consumers.Actors.KafkaConsumerActor`2[[System.Byte[], System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Byte[], System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]"
```

From chat:
> my consumer is basically a `KafkaConsumer.CommittableSource` that eventually produces messages using the `KafkaProducer.FlexiFlow<byte[], byte[], ICommittableOffset>( producerSettings)` and finally committing the offset with `Committer.Sink(committerSettings)`

Consumer config:
```
var consumerConfig = new ConsumerConfig
{
    EnableAutoCommit = true,
    EnableAutoOffsetStore = false,
    AllowAutoCreateTopics = true,
    AutoOffsetReset = AutoOffsetReset.Latest,
    ClientId = _streamingConfig.ConsumerId,
    SecurityProtocol = _streamingConfig.SecurityProtocol,
    SaslMechanism = _streamingConfig.SaslMechanism,
[...]

PR #254, opened 2021-11-16
Add default exception decider

Arkatufus:


src/Akka.Streams.Kafka/Akka.Streams.Kafka.csproj
```diff
@@ -5,6 +5,7 @@
     <Description>Apache Kafka adapter for Akka.NET Streams</Description>
     <TargetFramework>$(NetStandardLibVersion)</TargetFramework>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+      <LangVersion>9.0</LangVersion>
```
Aaronontheweb, line 8: Can't use 9.0 without upgrading the runtime I think.

Arkatufus: Downgraded to 8.0

src/Akka.Streams.Kafka/Dsl/KafkaConsumer.cs
```diff
@@ -42,7 +44,8 @@ public static class KafkaConsumer
         /// </summary>
         public static Source<ConsumeResult<K, V>, IControl> PlainExternalSource<K, V>(IActorRef consumer, IManualSubscription subscription)
         {
-            return Source.FromGraph(new ExternalPlainSourceStage<K, V>(consumer, subscription));
+            return Source.FromGraph(new ExternalPlainSourceStage<K, V>(consumer, subscription))
+                .AddAttributes(ActorAttributes.CreateSupervisionStrategy(new DefaultConsumerDecider<K, V>().Decide));
```
Aaronontheweb, line 48: Wrong place to do it. You do it here: https://github.com/akkadotnet/Akka.Streams.Kafka/blob/4f4228199fdacb83d772367c5328812cf9f65357/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs#L42

Aaronontheweb: Fix this everywhere.

Aaronontheweb (changes requested):
Need to move supervision to the right place.

Need to test that this code does the right thing in some of your reproduction scenarios.

Need to validate that this supervision strategy can be overridden from user code.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs
```diff
@@ -27,19 +28,25 @@ public abstract class KafkaSourceStage<K, V, TMessage> : GraphStageWithMateriali
         
         /// <inheritdoc />
         public override SourceShape<TMessage> Shape { get; }
+        
+        protected bool AutoCreateTopics { get; }
```
Aaronontheweb, line 32: Would this get set to true per this setup? https://github.com/akkadotnet/Akka.Streams.Kafka/issues/228

https://github.com/Aaronontheweb/Akka.Streams.Kafka.Reproduction/blob/75d88a432b747437dd1c4b20fa30e3d766727294/src/Akka.Streams.Kafka.Reproduction.Consumer/Program.cs#L33-L42

And 

https://github.com/Aaronontheweb/Akka.Streams.Kafka.Reproduction/blob/75d88a432b747437dd1c4b20fa30e3d766727294/src/Akka.Streams.Kafka.Reproduction.Consumer/Program.cs#L73

i.e. the `WithProperty` method setting this to `true`?

src/Akka.Streams.Kafka/Dsl/KafkaConsumer.cs
Arkatufus: This does not work, InitialAttributes is somehow injected to the end of the attribute list, not the first.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs
Arkatufus: No

Aaronontheweb: That might be an issue then. Need to detect that no matter how the user sets it.

Aaronontheweb: Is there a way we can check via the underlying Kafka client itself rather than our settings class?

Aaronontheweb: (What's the best way to do this?)

src/Akka.Streams.Kafka/Supervision/DefaultDecider.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+        
+        public Directive Decide(Exception e)
+        {
+            switch (e)
+            {
+                case ConsumeException ce:
+                    if (ce.Error.IsFatal)
+                        return Directive.Stop;
+                    if (ce.Error.Code == ErrorCode.UnknownTopicOrPart && _autoCreateTopics)
```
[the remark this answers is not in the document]
Arkatufus: Done

src/Akka.Streams.Kafka/Supervision/DefaultProducerDecider.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using Akka.Actor;
+using Akka.Streams.Kafka.Settings;
+using Confluent.Kafka;
+
+namespace Akka.Streams.Kafka.Supervision
+{
+    public class DefaultProducerDecider<TKey, TValue>
```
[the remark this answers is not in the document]
Arkatufus: Done

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs
Arkatufus: Code moved. The code will try its best to look up this value from the ConsumerSettings, the only exception is when you use external source actor, since it would not have access to the original settings.

Aaronontheweb: We can document that edge case.

Aaronontheweb:
Is this ready for review?

Arkatufus:
Yes

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -44,15 +45,19 @@ internal abstract class BaseSingleSourceLogic<K, V, TMessage> : GraphStageLogic
...
+            bool autoCreateTopics) 
             : base(shape)
         {
             _shape = shape;
             _messageBuilder = messageBuilderFactory(this);
             Control = new BaseSingleSourceControl(_shape, Complete, SetKeepGoing, GetAsyncCallback, PerformShutdown);
             
-            var supervisionStrategy = attributes.GetAttribute<ActorAttributes.SupervisionStrategy>(null);
+            // TODO: Move this to the GraphStage.InitialAttribute when it is fixed (https://github.com/akkadotnet/akka.net/issues/5388)
+            var supervisionStrategy = attributes.GetAttribute(new ActorAttributes.SupervisionStrategy(new DefaultConsumerDecider(autoCreateTopics).Decide));
```
Aaronontheweb, line 60: Good, this is a reasonable way of doing it

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -123,11 +128,18 @@ protected virtual void MessageHandling((IActorRef, object) args)
                 
                 case Status.Failure failure:
                     var exception = failure.Cause;
-                    if (_decider(exception) == Directive.Stop)
+                    var cause = exception is KafkaException ke ? ke.Error.Reason : exception.Message; 
+                    var directive = _decider(exception); 
+                    if (directive == Directive.Stop)
```
Aaronontheweb, line 133: What happens if we get a `Directive.Restart`, which is the default?

Aaronontheweb: (default for supervision strategies in general)

src/Akka.Streams.Kafka/Supervision/DefaultConsumerDecider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Akka.Streams.Kafka.Supervision
+{
+    /// <summary>
+    /// Default Supervision decider used internally inside all consumer source stages.
+    ///
+    /// You can extend and override the virtual methods in this class to provide your own
+    /// custom supervision decider. 
+    /// </summary>
+    public class DefaultConsumerDecider
```
Aaronontheweb, line 14: LGTM

Aaronontheweb (commented):
Left some questions for you

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: Resume and Restart is pretty much the same at this time. Since the SourceActor is hosed every time it encounters an error, we basically have to restart the stream

Arkatufus: ie, clear the stage buffer and recreate a new KafkaConsumerActor

Arkatufus: Oh, i think i forgot something...

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -123,15 +128,28 @@ protected virtual void MessageHandling((IActorRef, object) args)
...
+                    }
+
+                    if(Log.IsInfoEnabled)
+                        Log.Info(exception, "Source stage failure [{0}] handled with Supervision Directive [{1}]", cause, directive);
                     
                     var isSerializationError = exception is ConsumeException cEx && cEx.Error.IsSerializationError();
                     if (isSerializationError)
-                        break;
+                    {
+                        if (directive == Directive.Resume)
```
Arkatufus, line 147: Need to make sure that the user have an option to restart when serialization error occured

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
Arkatufus: Added code to check if the decider asked the stage to resume or restart when a serialization error happened

Aaronontheweb (approved):
Crash on escalate directives in both stages; contribute error-handling documentation to README (make it easy to find, include code samples - esp for customizing error-handling and explaining the defaults) in separate PR.

PR #262, opened 2021-11-19
Fix Kafka polling method

Arkatufus:
On previous polling implementation, when an exception happened during a poll, all of the messages consumed before the exception was thrown were discarded. The last offset/message pushed into the stream can be different, depending on how fast the broker and client exchanges messages and when the polling timer ran.

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
+++ b/src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
@@ -124,7 +124,7 @@ namespace Akka.Streams.Kafka.Tests
         // In this test, the exception happened inside the consumer source while it is deserializing the value
         // Since we're restarting the stream, the output should be gapless
         [Fact]
-        public async Task SupervisionStrategy_Restart_Decider_on_Consumer_should_be_gapless()
+        public async Task Directive_Restart_on_failed_Consumer_should_restart_Consumer()
         {
             var topic = CreateTopic(1);
             var group = CreateGroup(1);
@@ -162,20 +162,27 @@ namespace Akka.Streams.Kafka.Tests
             
             var (_, probe) = KafkaConsumer
                 .PlainSource(consumerSettings, Subscriptions.Assignment(topicPartition))
-                .Select(c => c.Message.Value)
                 .WithAttributes(ActorAttributes.CreateSupervisionStrategy(Decider))
+                .Select(c => c.Message.Value)
                 .ToMaterialized(this.SinkProbe<int>(), Keep.Both)
                 .Run(Materializer);
 
-            probe.Request(10);
-            for (var i = 1; i < 11; i++)
+            probe.Request(20);
+            var pulled = new List<int>();
+            for (var i = 0; i < 14; i++)
             {
-                probe.ExpectNext(i, TimeSpan.FromSeconds(10)); 
+                var msg = probe.ExpectNext();
+                pulled.Add(msg);
             }
+
+            probe.ExpectNoMsg(TimeSpan.FromSeconds(2));
             probe.Cancel();
+
+            pulled.Should().BeEquivalentTo(new[] {1, 2, 3, 4, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10});
             
-            callCount.Should().Be(1);
-            serializationCallCount.Should().Be(1);
+            // Decider should be called twice, because deciders are called in BaseSingleSourceLogic and KafkaConsumerActor 
+            callCount.Should().Be(2);
+            serializationCallCount.Should().Be(2);
         }        
         
         [Fact]
@@ -463,7 +470,8 @@ namespace Akka.Streams.Kafka.Tests
 
             probe.Request(elementsCount);
             probe.ExpectNoMsg(TimeSpan.FromSeconds(10));
-            callCount.Should().Be(elementsCount);
+            // this is twice elementCount because Decider is called twice on each exceptions
+            callCount.Should().Be(elementsCount*2);
             probe.Cancel();
         }        
         
@@ -491,12 +499,13 @@ namespace Akka.Streams.Kafka.Tests
 
             probe.Request(elementsCount);
             probe.ExpectNoMsg(TimeSpan.FromSeconds(10));
-            decider.CallCount.Should().Be(elementsCount);
+            // this is twice elementCount because Decider is called twice on each exceptions
+            decider.CallCount.Should().Be(elementsCount*2);
             probe.Cancel();
         }
         
         [Fact]
-        public async Task Default_Decider_on_PlainSource_should_stop_on_internal_error()
+        public async Task Default_Decider_on_PlainSource_should_resume_on_KafkaException()
         {
             int elementsCount = 10;
             var topic1 = CreateTopic(1);
@@ -508,19 +517,13 @@ namespace Akka.Streams.Kafka.Tests
             
             var settings = CreateConsumerSettings<Null, string>(group1).WithAutoCreateTopicsEnabled(false);
 
-            // Stage will fail because we're trying to subscribe to partition 5, which does not exist.
+            // Stage produce Error with ErrorCode.Local_UnknownPartition because we're trying to subscribe to partition 5, which does not exist.
             var probe = KafkaConsumer
                 .PlainSource(settings, Subscriptions.Assignment(new TopicPartition(topic1, 5)))
                 .Select(c => c.Value)
                 .RunWith(this.SinkProbe<string>(), Materializer);
 
             probe.Request(elementsCount);
-            var error = probe.ExpectEvent(TimeSpan.FromSeconds(5));
-            error.Should().BeOfType<TestSubscriber.OnError>();
-            var exception = ((TestSubscriber.OnError)error).Cause;
-            exception.Should().BeOfType<KafkaException>();
-            ((KafkaException) exception).Error.Code.Should().Be(ErrorCode.Local_UnknownPartition);
-
             probe.ExpectNoMsg(TimeSpan.FromSeconds(1));
             
             probe.Cancel();
--- a/src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
+++ b/src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
@@ -178,7 +178,7 @@ namespace Akka.Streams.Kafka.Tests
             probe.ExpectNoMsg(TimeSpan.FromSeconds(2));
             probe.Cancel();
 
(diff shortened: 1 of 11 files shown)


src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -545,22 +537,45 @@ private void Poll()
...
+        }
+
+        private (List<ConsumeResult<K, V>>, Exception) PollKafka(CancellationToken token)
         {
             ConsumeResult<K, V> consumed = null;
             var i = 10; // 10 poll attempts
             var timeout = Math.Max((int) _pollTimeout.TotalMilliseconds / i, 1);
-            var pooled = new List<ConsumeResult<K, V>>();
+            var polled = new List<ConsumeResult<K, V>>();
             do
```
Arkatufus, line 562: New polling method. Instead of discarding all the consumed message when an exception is thrown, we return both the polled messages before the exception and the exception. We then first push all of the consumed messages and then process the exception. This makes polling consistent on both slow and fast brokers and allows us to implement correct stage restart behavior.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -475,67 +477,57 @@ private void Poll()
...
+                var resumeThese = currentAssignment.Where(partitionsToFetch.Contains).ToList();
+                var pauseThese = currentAssignment.Except(resumeThese).ToList();
+                PausePartitions(pauseThese);
+                ResumePartitions(resumeThese);
+
+                using (var cts = new CancellationTokenSource(_settings.PollTimeout))
+                {
+                    var (polled, exception) = PollKafka(cts.Token);
+                    ProcessResult(partitionsToFetch, polled);
+                    ProcessExceptions(exception);
```
Arkatufus, line 527: Process all messages that happened before we hit any exceptions, then process the exception, if any.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -545,22 +537,45 @@ private void Poll()
             }
         }
 
-        private List<ConsumeResult<K, V>> PollKafka(CancellationToken token)
+        private void ProcessExceptions(Exception exception)
```
Arkatufus, line 540: All exceptions are routed here, we don't kill the actor on all errors anymore, we keep it alive if the Decider returns Directive.Resume.

Arkatufus: The side effect of this is the Decider delegate gets called twice, inside the KafkaConsumerActor and inside the stage itself.

Aaronontheweb:
Wow, looks like a big change! I know we had discussed this but had no idea you were getting to work on it right away. Good stuff!

Arkatufus:
I'm wondering if thrown exceptions should be emitted on the next polling cycle instead of right away, to let the stream catch up with the messages, if it were to fail instead of resume/restart.

Arkatufus:
@Aaronontheweb Yeah, this is required to fix one of the racy specs, if not both.

Arkatufus:
Plus I need to update some of the specs since we changed the default stream behavior, rechecked that they should be doing what they were supposed to be doing

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -323,14 +332,7 @@ private void ApplySettings(ConsumerSettings<K, V> updatedSettings)
 
                 var localSelf = Self;
                 _consumer = _settings.CreateKafkaConsumer(
-                    consumeErrorHandler: (c, e) =>
```
Aaronontheweb, line 326: Interesting - so we just deliver a message to ourselves and handle it inside the actor's scope rather than externally now? Seems like a good idea, but why this change? Is it just to make sure that the decider is "in charge?"

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: Only gets called twice in the event that the actor propagates the exception, right?

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -545,22 +537,45 @@ private void Poll()
...
+                catch (Exception e)
+                {
+                    return (polled, e);
+                }
                 if (consumed != null)
-                    pooled.Add(consumed);
+                    polled.Add(consumed);
                 i--;
-            } while (i > 0 && !token.IsCancellationRequested);
+            } while (i > 0 && consumed != null && !token.IsCancellationRequested);
```
Aaronontheweb, line 576: So this will exit right away as soon as we receive a single message - no point in building a list of them if that's the case, no?

Aaronontheweb (commented):
Left some questions

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: No, Consumer.Consume will return a null if its internal message queue is empty. So instead of waiting for the whole timeout duration, we just bail out of the polling method and return the messages that we polled up to that point.

Arkatufus: You can consider the Consumer.Consume returning null the same as C string null terminator

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs
```diff
@@ -38,10 +38,12 @@ protected KafkaSourceStage(string stageName)
         {
             StageName = stageName;
             Shape = new SourceShape<TMessage>(Out);
+            InitialAttributes = Attributes.CreateName(StageName)
+                .And(ActorAttributes.CreateSupervisionStrategy(new DefaultConsumerDecider(true).Decide));
         }
 
         /// <inheritdoc />
-        protected override Attributes InitialAttributes => Attributes.CreateName(StageName);
+        protected override Attributes InitialAttributes { get; }
```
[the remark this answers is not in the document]
Arkatufus: I was thinking more of a single instantiation, instead of having the attributes getting instantiated every time get_InitialAttributes is called.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: Yes, the consumeErrorHandler callback is called on internal errors like unknown topic or partition. Not all errors are propagated to the `Consume` method. By routing this error to the actor, error handling become a lot more consistent, Plus we avoid any possible threading problems.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -475,67 +477,57 @@ private void Poll()
...
+                ResumePartitions(resumeThese);
+
+                using (var cts = new CancellationTokenSource(_settings.PollTimeout))
+                {
+                    var (polled, exception) = PollKafka(cts.Token);
+                    ProcessResult(partitionsToFetch, polled);
+                    ProcessExceptions(exception);
+                }
             }
-            // Workaroud for https://github.com/confluentinc/confluent-kafka-dotnet/issues/1366
```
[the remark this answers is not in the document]
Arkatufus: This is handled inside the decider now, that way it can actually be overridden if the user wishes to.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: For now, it is always processed twice. It can be changed so that it doesn't propagate Resume directived errors.
I always propagate it because logging is handled by the stage, I've tried to make the actor as dumb as possible and hand over all responsibilities to the stage.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -162,20 +162,27 @@ await Source.From(Enumerable.Range(1, 10))
...
             {
-                probe.ExpectNext(i, TimeSpan.FromSeconds(10)); 
+                var msg = probe.ExpectNext();
+                pulled.Add(msg);
             }
+
+            probe.ExpectNoMsg(TimeSpan.FromSeconds(2));
             probe.Cancel();
+
+            pulled.Should().BeEquivalentTo(new[] {1, 2, 3, 4, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10});
```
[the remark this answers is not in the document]
Arkatufus: That... I'm not sure off, I'll come back to you.

src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -162,20 +162,27 @@ await Source.From(Enumerable.Range(1, 10))
...
             {
-                probe.ExpectNext(i, TimeSpan.FromSeconds(10)); 
+                var msg = probe.ExpectNext();
+                pulled.Add(msg);
             }
+
+            probe.ExpectNoMsg(TimeSpan.FromSeconds(2));
             probe.Cancel();
+
+            pulled.Should().BeEquivalentTo(new[] {1, 2, 3, 4, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, opt => opt.WithStrictOrdering());
```
Arkatufus, line 181: This now checks for strict ordering

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/KafkaSourceStage.cs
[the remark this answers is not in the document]
Aaronontheweb: Ah ok

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Aaronontheweb: Ok got it - I read the boolean logic backwards there

Aaronontheweb (approved):
LGTM - can merge when out of draft

Aaronontheweb:
All of the partitioned source specs are failing @Arkatufus

PR #284, opened 2022-03-15
Repro CommittablePartitionedSource generates hung Kafka consumer

Arkatufus:
This is NOT a spec, this code is ment to hammer CommittablePartitionedSource to check for bugs.


Aaronontheweb:
I think this has been fixed - was probably the partitions lost issue we fixed late in 2024.

PR #406, opened 2024-09-26
fix partition sample

Aaronontheweb:


examples/Kafka.Partitioned.Consumer/Actors/ConsumerWorkerActor.cs
```diff
@@ -54,7 +54,7 @@ protected override void PreStart()
             _control = KafkaConsumer.CommittableSource(_settings, _subscription)
                 .Ask<ICommittable>(Self, TimeSpan.FromSeconds(1), 1)
                 .ToMaterialized(Committer.Sink(committerDefaults), DrainingControl<NotUsed>.Create)
-                .Run(Context.System.Materializer());
+                .Run(Context.Materializer());
```
Aaronontheweb, line 57: This will leak streams actors under the current design - but moving them as children of this actor will fix it

PR #410, opened 2024-10-10
Fix possible TransactionalSourceLogic partition drain bug

Arkatufus:


src/Akka.Streams.Kafka/Stages/Consumers/Abstract/TransactionalSourceLogic.cs
```diff
@@ -165,7 +165,8 @@ private bool WaitForDraining(IImmutableSet<TopicPartition> partitions)
         {
             try
             {
-                StageActor.Ref.Ask(new Drain(partitions, Option<IActorRef>.None, new Drained()), _settings.CommitTimeout);
+                StageActor.Ref.Ask(new Drain(partitions, Option<IActorRef>.None, new Drained()), _settings.CommitTimeout)
```
Aaronontheweb, line 168: lol makes sense - have to ACTUALLY wait for it to drain, not just hope it does

Aaronontheweb (approved):
LGTM

PR #418, opened 2025-01-16
Refactor KafkaConsumerActor event handling from Mailbox to direct method call

Arkatufus:
Possible fix for #414 and #415

## Changes

* Refactor KafkaConsumerActor event handling from using `Self.Tell()` to direct method call
* Implement missing OnPartitionsLost handler

PR #431, opened 2025-02-28
Revert #418

Arkatufus:
## Changes

Revert #418, make partition events processing back to use `Tell()`

### DIFF
--- a/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
+++ b/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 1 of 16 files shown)


src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
```diff
@@ -43,7 +43,7 @@ public static IImmutableSet<T> ToImmutableSet<T>(this IEnumerable<T> collection)
         /// <summary>
         /// Split a list into two list depending on the boolean return value of the predicate.
         /// </summary>
-        public static (IImmutableList<T> True, IImmutableList<T> False) Partition<T>(this IImmutableList<T> list, Predicate<T> predicate)
+        public static (List<T> True, List<T> False) Partition<T>(this List<T> list, Predicate<T> predicate)
```
Aaronontheweb, line 46: Why do the lists need to be mutable?

Aaronontheweb (dismissed):
Needs a self-review

Issue #415, opened 2025-01-07
System.ArgumentException: Unexpected records polled potentially thrown during a rebalance


jblackburn21:
**Version Information**
Version of Akka.NET? 1.5.33
Which Akka.NET Modules? Akka.Streams.Kafka

**Describe the bug**
We use the `KafkaConsumer.CommittablePartitionedSource` running in kubernetes with autoscaling enabled. When Pods are spun up/down, it is possible for the `SubSourceLogic` to get out of sync with the `TopicPartition` assignment on the consumer, which results in an ex being thrown and the consumer shutting down. The `PartitionsAssignedHandler` and `PartitionsRevokedHandler` handlers are a side effect of `_consumer.Consume()`, and doesn't appear handle timing properly.

**To Reproduce**

**NOTE:** I added additional logging to `KafkaConsumerActor` in order to better capture current start during rebalancing

1. Run kafka locally using docker
2. Create a topic with 10 partitions, in this case I'm using a `members` topic
3. Run a producer so that messages are generated with a key to produce to all partitions
4. Start 3 instances of a consumer:

    - Start Consumer 1, all 10 partitions are assigned
    
    ```bash
    consumer 1 - 10 assignments: members [[0]], members [[1]], members [[2]], members [[3]], members [[4]], members [[5]], members [[6]], members [[7]], members [[8]], members [[9]]
    ```

    - Start Consumer 2, partitions are reassigned to the new consumer

    ```bash
    consumer 1 - 5 assignments: members [[0]], members [[1]], members [[2]], members [[3]], members [[4]]
    consumer 2 - 5 assignments: members [[5]], members [[6]], members [[7]], members [[8]], members [[9]]
    ```

    - Start Consumer 3, partitions are reassigned to the new consumer

    ```bash
    consumer 1 - 4 assignments: members [[0]], members [[1]], members [[2]], members [[3]]
    consumer 2 - 3 assignments: members [[4]], members [[5]], members [[6]]
    consumer 3 - 3 assignments: members [[7]], members [[8]], members [[9]]
    ```

5. Stop Consumer 1, check logs for consumers 2 and 3 to see if they stopped. It may take multiple restarts of consumer 1 to trigger the error.

**Expected behavior**
The `KafkaConsumerActor` properly update the `_requests` and `_requestors` on `PartitionsRevokedHandler` so that it isn't expected to consume messages from partitions that are not longer assigned.

**Actual behavior**

[...]

PR #443, opened 2025-03-04
[WIP] rebalance integration specs

Aaronontheweb:
Reproduction for #415


Aaronontheweb:
Seeing a lot of these failures too, which we also saw in our test lab:

```
Stream failed due to One or more errors occurred. (Kafka commit took longer than: 00:00:15)
   at Akka.Streams.Kafka.Tests.Integration.RebalanceIntegrationTests.<>c__DisplayClass5_0.<<ShouldReBalanceWithoutArgumentExceptions>g__KillAndRelaunchFirstConsumer|0>d.MoveNext() in C:\Repositories\olympus\Akka.Streams.Kafka\src\Akka.Streams.Kafka.Tests\Integration\RebalanceIntegrationTests.cs:line 122
```

Aaronontheweb:
> Seeing a lot of these failures too, which we also saw in our test lab:
> 
> ```
> Stream failed due to One or more errors occurred. (Kafka commit took longer than: 00:00:15)
>    at Akka.Streams.Kafka.Tests.Integration.RebalanceIntegrationTests.<>c__DisplayClass5_0.<<ShouldReBalanceWithoutArgumentExceptions>g__KillAndRelaunchFirstConsumer|0>d.MoveNext() in C:\Repositories\olympus\Akka.Streams.Kafka\src\Akka.Streams.Kafka.Tests\Integration\RebalanceIntegrationTests.cs:line 122
> ```

Pretty sure this issue is related to how we handle commit stashing during a rebalance

src/Akka.Streams.Kafka.Tests/Integration/RebalanceIntegrationTests.cs
```diff
@@ -87,7 +88,11 @@ public async Task ShouldReBalanceWithoutArgumentExceptions()
         var killSwitch3 = CreateKillableStream(topic, settings, probe1.Ref, "consumer3");
 
         // make sure all 10 messages got processed
-        var msgs1 = await probe1.ReceiveNAsync(10).Cast<ConsumeResult<Null, string>>().ToListAsync();
+        await foreach (var msg in probe1.ReceiveNAsync(10))
+        {
```
Aaronontheweb, line 93: Not necessary - the previous code would blow up anyway if a `StreamFailed` message arrived

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -92,6 +92,8 @@ internal class KafkaConsumerActor<K, V> : ActorBase, ILogReceive, IWithTimers
         /// </summary>
         private IImmutableList<IActorRef> _rebalanceCommitSenders = ImmutableArray<IActorRef>.Empty;
 
+        private ImmutableList<TopicPartition> _pausedPartitions = ImmutableList<TopicPartition>.Empty;
```
Arkatufus, line 95: Fix: holder for newly assigned partitions

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -470,6 +481,8 @@ private void Poll()
                     var consumed = _consumer.Consume(0);
                     if (consumed != null)
                         throw new IllegalActorStateException("Consumed message should be null");
+                    PausePartitions(_pausedPartitions);
+                    _pausedPartitions = ImmutableList<TopicPartition>.Empty;
```
Arkatufus, line 485: Fix: 

It is not possible to pause newly assigned partitions inside `OnPartitionsAssigned` event callback; `ConfluentKafka.ConsumerClient` will throw because the callback was invoked _before_ the internal partitions assignment list is updated. This is a hack to circumvent that, we call `Pause()` directly after the `Consume()` invocation.

Note that this is safe because everything is still running in the same thread.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -460,7 +467,11 @@ private void Poll()
             var currentAssignment = _consumer.Assignment.ToImmutableList();
             var initialRebalanceInProcess = _rebalanceInProgress.Value;
 
-            if (_requests.IsEmpty())
+            var partitionsToFetch = _requests.Values.SelectMany(v => v.Topics)
+                .Where(p => currentAssignment.Contains(p))
+                .ToImmutableHashSet();
```
Arkatufus, line 472: Actual fix:

Before, we did not filter the expected fetch partitions against the currently assigned partitions in the `ConsumerClient`, this causes the "Unexpected records" bug.

Arkatufus (commented):
Self-review

PR #446, opened 2025-03-06
WIP: track down source of `IllegalActorStateException` messages

Aaronontheweb:


Aaronontheweb:
@Arkatufus and I suspected it, so I had to ask the Confluent folks if `Pause` / `Resume` are asynchronous - it looks like they are not: https://github.com/confluentinc/confluent-kafka-dotnet/issues/2418

Issue #447, opened 2025-03-06
Partition Rebalance + Data Consistency Issues


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.5.38 and earlier
Which Akka.NET Modules? Akka.Streams.Kafka

**Describe the bug**

A continuation of https://github.com/akkadotnet/Akka.Streams.Kafka/issues/415 and https://github.com/akkadotnet/Akka.Streams.Kafka/pull/446, the basic issue is that the way https://github.com/confluentinc/confluent-kafka-dotnet/ works is inherently different from the Confluent Java driver that the original Scala code for alpakka-kafka was written against.

Some key differences:

1. Java client works in batch, .NET client consumes one message at a time;
2. Java client has the ability to call `Pause` inside the partition assignment handler, whereas with the .NET client this throws a rather violent exception. That means we can't easily pause data from new partitions without some of the creative workarounds like the type @Arkatufus employed in https://github.com/akkadotnet/Akka.Streams.Kafka/pull/443
3. Even if we do call `.Pause` for all of the correct partitions right before we call `Consume` - there might _already_ be messages from partitions that have been assigned but not requested by Akka.Streams stages waiting in the buffer. This is one of the two edge cases that could trigger #415 
4. The flip side of bullet 3 is old messages from recently revoked partitions still sitting in the buffer. Based on our logs it looks like this error doesn't happen very much at all - it's almost entirely category 3 that's the issue.

**To Reproduce**

We have the spec we added in https://github.com/akkadotnet/Akka.Streams.Kafka/pull/443 but I think that needs to be hardened.

**Expected behavior**

Rebalancing must "just work"

**Actual behavior**

Running v1.5.38 in our test lab environment we see regular errors occur every few hours or so - that's with constant rebalancing of our partitions at least once every 90 seconds.

**Additional context**

So one sticky wicket is the behavior of the Confluent.Kafka driver in .NET itself - which appears to keep a local buffer of messages from partitions we didn't get to pause quickly enough or partitions that were recently revoked: https://github.com/confluentinc/confluent-kafka-dotnet/issues/2418

This is going to require us to RELAX how we handle some of the partition validation inside the `KafkaConsumerActor` in particular.

Aaronontheweb:
[...]

PR #449, opened 2025-03-07
Enable nullability

Aaronontheweb:
More cleanup before we implement https://github.com/akkadotnet/Akka.Streams.Kafka/issues/447


Aaronontheweb:
Could of breaking-sh changes here:

* `ConsumerSettings<K,V>` is now a `record` - but all of the old methods are still supported
* `ProducerSettings<K,V>` is now a `record` - and ditto

PR #452, opened 2025-03-11
Refactor `SubSourceStageLogic`; filter messages from revoked partitions in partitioned stream sources

### LINKED ISSUE #447: Partition Rebalance + Data Consistency Issues

Aaronontheweb:
## Changes

Numerous bug fixes and consistency issue fixes here. #415 still has not been addressed but I'm working my way towards that.

partially resolves #447 

## Checklist

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/RebalanceIntegrationTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/RebalanceIntegrationTests.cs
@@ -118,7 +118,7 @@ public class RebalanceIntegrationTests : KafkaIntegrationTests
         // Assert.True(control2.IsShutdown.IsCompleted);
     }
 
-    [Fact(Skip = "Filtering is not enabled on sub-sources just yet")]
+    [Fact]
     public async Task FetchedRecords_must_be_removed_from_the_partitioned_source_stage_when_a_partition_is_revoked()
     {
         const int count = 20;
--- a/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
+++ b/src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
@@ -6,7 +6,7 @@ using System.Linq;
 
 namespace Akka.Streams.Kafka.Extensions
 {
-    public static class CollectionExtensions
+    internal static class CollectionExtensions
     {
         /// <summary>
         /// Joins elements of the collection with given separator to single string
--- a/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
+++ b/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
@@ -44,13 +44,25 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         public static readonly ISubSourceCancellationStrategy Instance = new DoNothing();
         private DoNothing() {}
     }
+
+    /// <summary>
+    /// INTERNAL API
+    /// </summary>
+    internal static class SubSourceLogicHelpers
+    {
+        public sealed class CloseRevokedPartitions
+        {
+            public static readonly CloseRevokedPartitions Instance = new();
+            private CloseRevokedPartitions(){}
+        }
+    }
     
     /// <summary>
     /// Stage logic used to produce sub-sources per topic partitions
     /// </summary>
     internal class SubSourceLogic<K, V, TMessage> : TimerGraphStageLogic
     {
-        private class CloseRevokedPartitions { }
+        
 
         private readonly SourceShape<(TopicPartition, Source<TMessage, NotUsed>)> _shape;
         private readonly ConsumerSettings<K, V> _settings;
@@ -206,7 +218,7 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
 
         protected override void OnTimer(object timerKey)
         {
-            if (timerKey is CloseRevokedPartitions)
+            if (timerKey is SubSourceLogicHelpers.CloseRevokedPartitions)
             {
                 Log.Debug("#{0} Closing SubSources for revoked partitions: {1}", _actorNumber, _partitionsToRevoke.JoinToString(", "));
 
@@ -298,7 +310,7 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         {
             _partitionsToRevoke = _partitionsToRevoke.Union(revoked.Select(r => r.TopicPartition));
 
-            ScheduleOnce(new CloseRevokedPartitions(), _settings.WaitClosePartition);
+            ScheduleOnce(SubSourceLogicHelpers.CloseRevokedPartitions.Instance, _settings.WaitClosePartition);
         }
 
         private void HandleSubsourceCancelled((TopicPartition, ISubSourceCancellationStrategy) obj)
--- a/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
+++ b/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Akka.Actor;
-using Akka.Event;
 using Akka.Streams.Dsl;
 using Akka.Streams.Kafka.Extensions;
 using Akka.Streams.Kafka.Helpers;
@@ -30,9 +29,17 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         public sealed class CloseRevokedPartitions
         {
             public static readonly CloseRevokedPartitions Instance = new();
-            private CloseRevokedPartitions(){}
+
+            private CloseRevokedPartitions()
+            {
+            }
         }
-        
+
+        /// <summary>
+        /// Used to determine how the <see cref="SubSourceLogic{K,V,TMessage}"/> will handle the cancellation of
+        /// a sub source stage. The default behavior requested by the <see cref="SubSourceStageLogic{K,V,TMessage}"/> is to ask
+        /// the consumer to seek to the last committed offset and then re-emit the sub source stage downstream.
+        /// </summary>
         internal interface ISubSourceCancellationStrategy;
 
         internal sealed record SeekToOffsetAndReEmit(long Offset) : ISubSourceCancellationStrategy;
@@ -40,28 +47,61 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         internal sealed class ReEmit : ISubSourceCancellationStrategy
         {
             public static readonly ISubSourceCancellationStrategy Instance = new ReEmit();
-            private ReEmit() {}
+
+            private ReEmit()
+            {
+            }
         }
 
         internal sealed class DoNothing : ISubSourceCancellationStrategy
         {
             public static readonly ISubSourceCancellationStrategy Instance = new DoNothing();
-            private DoNothing() {}
+
+            private DoNothing()
+            {
+            }
+        }
+
+        /// <summary>
+        /// INTERNAL API
+        /// </summary>
+        /// <param name="Control">Control for SubSourceStageLogic</param>
+        /// <param name="StageActor">Actor for the SubSourceStageLogic</param>
+        public sealed record ControlAndStageActor(IControl Control, IActorRef StageActor);
+
+        public sealed record SubSourceStageLogicControl(
+            TopicPartition TopicPartition,
+            ControlAndStageActor ControlAndStageActor,
+            Action<IImmutableSet<TopicPartitionOffset>> FilterRevokedPartitionsCb);
+
+        /// <summary>
+        /// Factory method to create a <see cref="SubSourceStageLogic{K,V,TMessage}"/> within
+        /// <see cref="SubSourceLogic{K,V,TMessage}"/> where the context parameters exist.
+        /// </summary>
+        public interface ISubSourceStageLogicFactory<K, V, TMessage>
+        {
+            SubSourceStageLogic<K, V, TMessage> Create(SourceShape<TMessage> shape, TopicPartition tp,
+                IActorRef consumerActor,
+                Action<SubSourceStageLogicControl> subSourceStartedCb,
+                Action<(TopicPartition partition, ISubSourceCancellationStrategy cancellationStrategy)>
+                    subSourceCancelledCb,
+                int actorNumber);
         }
     }
-    
+
     /// <summary>
     /// Stage logic used to produce sub-sources per topic partitions
     /// </summary>
     internal class SubSourceLogic<K, V, TMessage> : TimerGraphStageLogic
     {
-        
-
         private readonly SourceShape<(TopicPartition, Source<TMessage, NotUsed>)> _shape;
         private readonly ConsumerSettings<K, V> _settings;
         private readonly IAutoSubscription _subscription;
         private readonly IMessageBuilder<K, V, TMessage> _messageBuilder;
-        private readonly Option<Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>>> _getOffsetsOnAssign;
+
+        private readonly Option<Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>>>
+            _getOffsetsOnAssign;
+
         private readonly Action<IImmutableSet<TopicPartition>> _onRevoke;
 
         private readonly int _actorNumber = KafkaConsumerActorMetadata.NextNumber();
@@ -70,7 +110,10 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         private readonly Action<IImmutableSet<TopicPartitionOffset>> _partitionRevokedCallback;
         private readonly Action<(TopicPartition, ISubSourceCancellationStrategy)> _subsourceCancelledCallback;
         private readonly Action<(TopicPartition, IControl)> _subsourceStartedCallback;
-        private readonly Action<(IImmutableSet<TopicPartition>, IImmutableSet<TopicPartitionOffset>)> _offsetsFromExternalResponseCb;
+
+        private readonly Action<(IImmutableSet<TopicPartition>, IImmutableSet<TopicPartitionOffset>)>
+            _offsetsFromExternalResponseCb;
+
         private readonly Action<ConsumerFailed> _stageFailCallback;
         private readonly Decider _decider;
 
@@ -83,7 +126,9 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         /// We have created a source for these partitions, but it has not started up and is not in subSources yet.
         /// </summary>
         private IImmutableSet<TopicPartition> _partitionsInStartup = ImmutableHashSet<TopicPartition>.Empty;
-        private IImmutableDictionary<TopicPartition, IControl> _subSources = ImmutableDictionary<TopicPartition, IControl>.Empty;
+
+        private IImmutableDictionary<TopicPartition, IControl> _subSources =
+            ImmutableDictionary<TopicPartition, IControl>.Empty;
 
         /// <summary>
         /// Kafka has signalled these partitions are revoked, but some may be re-assigned just after revoking.
@@ -98,10 +143,12 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
         /// <summary>
         /// SubSourceLogic
         /// </summary>
-        public SubSourceLogic(SourceShape<(TopicPartition, Source<TMessage, NotUsed>)> shape, ConsumerSettings<K, V> settings,
-                              IAutoSubscription subscription, Func<SubSourceLogic<K, V, TMessage>, IMessageBuilder<K, V, TMessage>> messageBuilderFactory,
-                              Option<Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>>> getOffsetsOnAssign,
-                              Action<IImmutableSet<TopicPartition>> onRevoke, Attributes attributes)
+        public SubSourceLogic(SourceShape<(TopicPartition, Source<TMessage, NotUsed>)> shape,
+            ConsumerSettings<K, V> settings,
+            IAutoSubscription subscription,
+            Func<SubSourceLogic<K, V, TMessage>, IMessageBuilder<K, V, TMessage>> messageBuilderFactory,
+            Option<Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>>> getOffsetsOnAssign,
+            Action<IImmutableSet<TopicPartition>> onRevoke, Attributes attributes)
             : base(shape)
         {
             _shape = shape;
@@ -114,19 +161,24 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
             var supervisionStrategy = attributes.GetAttribute<ActorAttributes.SupervisionStrategy>();
             _decider = supervisionStrategy != null ? supervisionStrategy.Decider : Deciders.StoppingDecider;
 
-            Control = new SubSourcePromiseControl(_shape, Complete, SetKeepGoing, GetAsyncCallback, GetAsyncCallback, PerformStop, PerformShutdown);
+            Control = new SubSourcePromiseControl(_shape, Complete, SetKeepGoing, GetAsyncCallback, GetAsyncCallback,
+                PerformStop, PerformShutdown);
 
-            _updatePendingPartitionsAndEmitSubSourcesCallback = GetAsyncCallback<IImmutableSet<TopicPartition>>(UpdatePendingPartitionsAndEmitSubSources);
+            _updatePendingPartitionsAndEmitSubSourcesCallback =
+                GetAsyncCallback<IImmutableSet<TopicPartition>>(UpdatePendingPartitionsAndEmitSubSources);
             _partitionAssignedCallback = GetAsyncCallback<IImmutableSet<TopicPartition>>(HandlePartitionsAssigned);
             _partitionRevokedCallback = GetAsyncCallback<IImmutableSet<TopicPartitionOffset>>(HandlePartitionsRevoked);
             _stageFailCallback = GetAsyncCallback<ConsumerFailed>(FailStage);
-            _subsourceCancelledCallback = GetAsyncCallback<(TopicPartition, ISubSourceCancellationStrategy)>(HandleSubsourceCancelled);
+            _subsourceCancelledCallback =
+                GetAsyncCallback<(TopicPartition, ISubSourceCancellationStrategy)>(HandleSubsourceCancelled);
             _subsourceStartedCallback = GetAsyncCallback<(TopicPartition, IControl)>(HandleSubsourceStarted);
-            _offsetsFromExternalResponseCb = GetAsyncCallback<(IImmutableSet<TopicPartition>, IImmutableSet<TopicPartitionOffset>)>(OffsetsFromExternalResponseCallback);
+            _offsetsFromExternalResponseCb =
+                GetAsyncCallback<(IImmutableSet<TopicPartition>, IImmutableSet<TopicPartitionOffset>)>(
+                    OffsetsFromExternalResponseCallback);
 
             SetHandler(shape.Outlet, onPull: EmitSubSourcesForPendingPartitions, onDownstreamFinish: PerformShutdown);
         }
-        
+
         protected void ConfigureSubscription(Action<IImmutableSet<TopicPartition>> partitionsAssignedCb,
             Action<IImmutableSet<TopicPartitionOffset>> partitionsRevokedCb)
         {
@@ -152,11 +204,13 @@ namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
 
             IPartitionEventHandler CreateRebalanceListener(IAutoSubscription subscription)
             {
-                return new PartitionEventHandlers.Chain(subscription.PartitionEventsHandler.GetOrElse(PartitionEventHandlers.Empty.Instance), new PartitionEventHandlers.AsyncCallbacks(subscription, SourceActor.Ref, partitionsAssignedCb,
-                    partitionsRevokedCb));
+                return new PartitionEventHandlers.Chain(
+                    subscription.PartitionEventsHandler.GetOrElse(PartitionEventHandlers.Empty.Instance),
+                    new PartitionEventHandlers.AsyncCallbacks(subscription, SourceActor.Ref, partitionsAssignedCb,
+                        partitionsRevokedCb));
             }
         }
-        
+
         /// <summary>
         /// Opportunity for subclasses to add their logic to the partition assignment callbacks.
         /// </summary>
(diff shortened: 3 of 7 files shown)


src/Akka.Streams.Kafka.Tests/Integration/RebalanceIntegrationTests.cs
```diff
@@ -118,7 +118,7 @@ public async Task FetchedRecords_must_be_removed_from_source_stage_buffer_when_p
         // Assert.True(control2.IsShutdown.IsCompleted);
     }
 
-    [Fact(Skip = "Filtering is not enabled on sub-sources just yet")]
+    [Fact]
     public async Task FetchedRecords_must_be_removed_from_the_partitioned_source_stage_when_a_partition_is_revoked()
```
Aaronontheweb, line 122: The step now works, seeing as how we've implemented our filtering `IPartitionHandler` correctly inside substreams.

src/Akka.Streams.Kafka/Extensions/CollectionExtensions.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Akka.Streams.Kafka.Extensions
 {
-    public static class CollectionExtensions
+    internal static class CollectionExtensions
```
Aaronontheweb, line 9: Made this `internal` - no good reason to expose these Util classes to the `public`.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -18,54 +16,103 @@
...
 
 namespace Akka.Streams.Kafka.Stages.Consumers.Abstract
 {
-    internal interface ISubSourceCancellationStrategy { }
-
-    internal sealed class SeekToOffsetAndReEmit : ISubSourceCancellationStrategy
+    /// <summary>
+    /// INTERNAL API
+    /// </summary>
+    internal static class SubSourceLogic
```
Aaronontheweb, line 26: Moved all of the key `SubSourceLogic` messages inside the static `SubSourceLogic` class

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -18,54 +16,103 @@
...
+            {
+            }
+        }
+
+        /// <summary>
+        /// INTERNAL API
+        /// </summary>
+        /// <param name="Control">Control for SubSourceStageLogic</param>
+        /// <param name="StageActor">Actor for the SubSourceStageLogic</param>
+        public sealed record ControlAndStageActor(IControl Control, IActorRef StageActor);
```
Aaronontheweb, line 69: State object for the control and the stage actor for a specific sub-source stage. One of these gets created per-partition.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -18,54 +16,103 @@
...
+
+        public sealed record SubSourceStageLogicControl(
```
Aaronontheweb, line 71: Bigger data object that contains the `ControlAndStageActor`, the `TopicPartition` owned by this stage, and the filter function for trimming messages from revoked partitions out of the applicable stages.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -18,54 +16,103 @@
...
+            TopicPartition TopicPartition,
+            ControlAndStageActor ControlAndStageActor,
+            Action<IImmutableSet<TopicPartitionOffset>> FilterRevokedPartitionsCb);
+
+        /// <summary>
+        /// Factory method to create a <see cref="SubSourceStageLogic{K,V,TMessage}"/> within
+        /// <see cref="SubSourceLogic{K,V,TMessage}"/> where the context parameters exist.
+        /// </summary>
+        public interface ISubSourceStageLogicFactory<K, V, TMessage>
```
Aaronontheweb, line 80: Factory method used to help inject the `SubSourceStageLogic` instances that are sub-classes inside things like the `PlainPartitionedSource` and the `CommittablePartitionedSource`

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -78,7 +125,9 @@ private class CloseRevokedPartitions { }
         /// We have created a source for these partitions, but it has not started up and is not in subSources yet.
         /// </summary>
         private IImmutableSet<TopicPartition> _partitionsInStartup = ImmutableHashSet<TopicPartition>.Empty;
-        private IImmutableDictionary<TopicPartition, IControl> _subSources = ImmutableDictionary<TopicPartition, IControl>.Empty;
+
+        private IImmutableDictionary<TopicPartition, SubSourceStageLogicControl> _subSources =
```
Aaronontheweb, line 129: We now store the `SubSourceStageLogicControl` type we defined earlier and associate it with its given `TopicPartition` - remember the `SubSourceLogic` is the parent object that emits the sub-streams, so this stage has to track and manage the lifecycle of all of them.

Aaronontheweb:
Got the following exception inside `PlainSource_consumes_messages_from_KafkaProducer_with_topicPartitionOffset_assignment`

```text
[INFO][03/11/2025 23:56:49.414Z][Thread 0029][client-group-1-09a50812-ede3-4983-91df-9bde9f187b7f(akka://PlainSourceIntegrationTests/user/StreamSupervisor-45)] Source stage failure [Local: Erroneous state] handled with Supervision Directive [Resume]
Cause: Confluent.Kafka.KafkaException: Local: Erroneous state
   at Confluent.Kafka.Impl.SafeKafkaHandle.Seek(String topic, Partition partition, Offset offset, Int32 millisecondsTimeout, Nullable`1 leaderEpoch)
   at Confluent.Kafka.Consumer`2.Seek(TopicPartitionOffset tpo)
   at Akka.Streams.Kafka.Stages.Consumers.Actors.KafkaConsumerActor`2.HandleSubscription(ISubscriptionRequest subscriptionRequest) in /home/vsts/work/1/s/src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs:line 415
```

Aaronontheweb:
> Got the following exception inside `PlainSource_consumes_messages_from_KafkaProducer_with_topicPartitionOffset_assignment`
> 
> ```
> [INFO][03/11/2025 23:56:49.414Z][Thread 0029][client-group-1-09a50812-ede3-4983-91df-9bde9f187b7f(akka://PlainSourceIntegrationTests/user/StreamSupervisor-45)] Source stage failure [Local: Erroneous state] handled with Supervision Directive [Resume]
> Cause: Confluent.Kafka.KafkaException: Local: Erroneous state
>    at Confluent.Kafka.Impl.SafeKafkaHandle.Seek(String topic, Partition partition, Offset offset, Int32 millisecondsTimeout, Nullable`1 leaderEpoch)
>    at Confluent.Kafka.Consumer`2.Seek(TopicPartitionOffset tpo)
>    at Akka.Streams.Kafka.Stages.Consumers.Actors.KafkaConsumerActor`2.HandleSubscription(ISubscriptionRequest subscriptionRequest) in /home/vsts/work/1/s/src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs:line 415
> ```

This was caused by calling `.Seek` immediately after calling `.Assign`: https://github.com/confluentinc/confluent-kafka-dotnet/issues/1303

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -89,39 +138,44 @@ private class CloseRevokedPartitions { }
...
+            Option<Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>>> getOffsetsOnAssign,
+            Action<IImmutableSet<TopicPartition>> onRevoke,
+            ISubSourceStageLogicFactory<K, V, TMessage> subSourceStageLogicFactory,
+            Attributes attributes)
             : base(shape)
         {
             _shape = shape;
             _settings = settings;
             _subscription = subscription;
-            _messageBuilder = messageBuilderFactory(this);
```
Aaronontheweb, line 105: the `IMessageBuilder` is now and internal implementation detail of the sub-stages, not the parent

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -147,23 +201,68 @@ protected void ConfigureSubscription(Action<IImmutableSet<TopicPartition>> parti
 
             IPartitionEventHandler CreateRebalanceListener(IAutoSubscription subscription)
             {
-                return new PartitionEventHandlers.Chain(subscription.PartitionEventsHandler.GetOrElse(PartitionEventHandlers.Empty.Instance), new PartitionEventHandlers.AsyncCallbacks(subscription, SourceActor.Ref, partitionsAssignedCb,
-                    partitionsRevokedCb));
+                return new PartitionEventHandlers.Chain(
+                    subscription.PartitionEventsHandler.GetOrElse(PartitionEventHandlers.Empty.Instance),
+                    new PartitionEventHandlers.AsyncCallbacks(subscription, SourceActor.Ref, partitionsAssignedCb,
```
Aaronontheweb, line 206: Works the same way that the `BaseSingleSourceStage` / `SingleSourceStage` both do - chains the `AsyncCallback` `IPartitionEventHandler` along with any user-defined ones.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -147,23 +201,68 @@ protected void ConfigureSubscription(Action<IImmutableSet<TopicPartition>> parti
...
+                        partitionsRevokedCb));
             }
         }
-        
+
+        private class FlushMessagesOfRevokedPartitionsHandler : IPartitionEventHandler
```
Aaronontheweb: Because we can't use default interface implementations, I had to copy and paste this code over from the `SingleSourceLogic` file and make some minor changes to this, but basically this is what is responsible for filtering out the messages from revoked partitions.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -147,23 +201,68 @@ protected void ConfigureSubscription(Action<IImmutableSet<TopicPartition>> parti
...
+            public void OnStop(IImmutableSet<TopicPartition> topicPartitions, IRestrictedConsumer consumer){}
+        }
+
         /// <summary>
         /// Opportunity for subclasses to add their logic to the partition assignment callbacks.
         /// </summary>
         protected virtual IPartitionEventHandler AddToPartitionAssignmentHandler(IPartitionEventHandler handler)
         {
-            return handler;
+            return new PartitionEventHandlers.Chain(handler, new FlushMessagesOfRevokedPartitionsHandler(this));
```
Aaronontheweb, line 250: Chains the `FlushMessagesOfRevokedPartitionsHandler` to the other two `IPartitionEventHandler`s

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -264,41 +375,34 @@ private void OffsetsFromExternalResponseCallback((IImmutableSet<TopicPartition>,
...
-            ConsumerActor.Ask(new KafkaConsumerActorMetadata.Internal.Seek(offsets), TimeSpan.FromSeconds(10))
-                .ContinueWith(t =>
+            _ = AskToSeekOffsets();
+            return;
+
+            async Task AskToSeekOffsets()
+            {
+                try
                 {
-                    if (t.IsCanceled)
```
Aaronontheweb, line 272: Simplified the error-handling here a great deal but just using a local `async Task` function and a `try...catch` block

Aaronontheweb:
There's definitely still something wrong with the manual offset-seeking implementations - and that's been the case before this PR too.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -331,25 +436,27 @@ private void HandleSubsourceCancelled((TopicPartition, ISubSourceCancellationStr
             }
         }
 
-        private void HandleSubsourceStarted((TopicPartition, IControl) obj)
+        private void HandleSubsourceStarted(SubSourceStageLogicControl sssLogicControl)
```
Aaronontheweb, line 431: Using the new data type here - this callback gets invoked once the substream starts

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -365,14 +472,14 @@ private void EmitSubSourcesForPendingPartitions()
                     _pendingPartitions = _pendingPartitions.Remove(topicPartition);
                     _partitionsInStartup = _partitionsInStartup.Add(topicPartition);
 
-                    var subSourceStage = new SubSourceStreamStage(
+                    var subSourceStage = new SubSourceStage<K,V,TMessage>(
```
Aaronontheweb, line 467: Each time we are given a new partition, we must emit a new `SubSourceStage` that belongs to that partition.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
             /// <inheritdoc />
             public override void PerformShutdown(Exception? ex) => _performShutdown(ex);
         }
+    }
 
-        private class SubSourceStreamStage : GraphStage<SourceShape<TMessage>>
```
Aaronontheweb, line 454: Moved this class outside of its parent - we had nested classes like 4 layers deep in this file.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
...
+            Out = new Outlet<TMessage>("out");
+            Shape = new SourceShape<TMessage>(Out);
+        }
 
-                Out = new Outlet<TMessage>("out");
-                Shape = new SourceShape<TMessage>(Out);
-            }
+        protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes)
+        {
+            return _subSourceStageLogicFactory.Create(Shape, _topicPartition, _consumerActor,
```
Aaronontheweb, line 601: The `SubSourceStage` is now basically uniform everywhere it gets used - the only thing that really changes is the `ISubSourceStageLogicFactory<K, V, TMessage>` implementation and its generic arguments.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
...
-            protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes)
+    /// <summary>
+    /// INTERNAL API
+    ///
+    /// A <see cref="SubSourceStageLogic{K,V,TMessage}"/> is the <see cref="GraphStageLogic"/>
+    /// of a <see cref="SubSourceStage{K,V,TMessage}"/>.
+    ///
+    /// This emits the actual Kafka messages downstream.
+    /// </summary>
+    internal abstract class SubSourceStageLogic<K, V, TMessage> : GraphStageLogic
```
Aaronontheweb, line 614: `SubSourceStageLogic` is where the real processing and interactivity with the `KafkaConsumerActor` happens. Most of the boilerplate is all wrapped inside this class but even still it should be subclassed for each flavor of stream stage it's going to be used with.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
...
+        private readonly SourceShape<TMessage> _shape;
+        private readonly TopicPartition _topicPartition;
+        private readonly IActorRef _consumerActor;
+        private readonly int _actorNumber;
+        private readonly IMessageBuilder<K, V, TMessage> _messageBuilder;
+        private readonly Action<SubSourceStageLogicControl> _subSourceStartedCallback;
+        private readonly KafkaConsumerActorMetadata.Internal.RequestMessages _requestMessages;
+        private bool _requested = false;
+        private StageActor _subSourceActor = null!;
+        private Queue<ConsumeResult<K, V>> _buffer = new();
```
Aaronontheweb, line 625: Just like with the non-partitioned stages: we never needed a `ConcurrentQueue<T>` - a regular old `Queue<T>` will work just fine.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
...
+        
+        public Action<IImmutableSet<TopicPartitionOffset>> FilterRevokedPartitionAsyncCallback =>
+            GetAsyncCallback<IImmutableSet<TopicPartitionOffset>>(FilterRevokedPartitions);
+        
+        private void FilterRevokedPartitions(IImmutableSet<TopicPartitionOffset> partitions)
```
Aaronontheweb, line 630: Filtering function that will asynchronously be invoked via the parent stage.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -450,191 +565,228 @@ public SubSourcePromiseControl(SourceShape<(
...
-                        case Status.Failure failure:
-                            FailStage(failure.Cause);
-                            break;
-                        case Terminated terminated when terminated.ActorRef.Equals(_consumerActor):
-                            FailStage(new ConsumerFailed());
-                            break;
-                    }
-                };
+            _subSourceStartedCallback(started);
+            _consumerActor.Tell(new KafkaConsumerActorMetadata.Internal.RegisterSubStage(_requestMessages.Topics),
```
Aaronontheweb, line 696: This is now implemented: we message the `KafkaConsumerActor` and let it know that we care about Topics X/Y/Z - that information will be stored inside of the `KafkaConsumerActor` to help ensure that messages don't get lost during rebalancing

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -56,13 +58,15 @@ internal class KafkaConsumerActor<K, V> : ActorBase, ILogReceive, IWithTimers
...
         /// </summary>
-        private IImmutableDictionary<IActorRef, KafkaConsumerActorMetadata.Internal.RequestMessages> _requests 
+        private IImmutableDictionary<IActorRef, KafkaConsumerActorMetadata.Internal.RequestMessages> _requests
             = ImmutableDictionary<IActorRef, KafkaConsumerActorMetadata.Internal.RequestMessages>.Empty;
-        
+
         /// <summary>
         /// Stores stage actors, requesting for more messages
         /// </summary>
-        private IImmutableSet<IActorRef> _requestors = ImmutableHashSet<IActorRef>.Empty;
```
Aaronontheweb, line 65: We upgraded from a hashset to a dictionary, which correlates the requested partitions to the requestor.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -187,80 +197,96 @@ protected override bool Receive(object message)
...
                     return true;
-                
+
                 case KafkaConsumerActorMetadata.Internal.ISubscriptionRequest subscribe:
                     HandleSubscription(subscribe);
                     return true;
-                
+
+                case KafkaConsumerActorMetadata.Internal.RegisterSubStage subStage:
+                    _stageActorsMap = _stageActorsMap.SetItem(subStage.TopicPartitions, Sender);
```
Aaronontheweb, line 213: Handle sub-stage registration

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -187,80 +197,96 @@ protected override bool Receive(object message)
...
+                    if (_settings.VerboseLogging)
+                        _log.Debug("Messages was requested, RequestId: {0}, Partitions: {1}", requestMessages.RequestId,
+                            string.Join(", ", requestMessages.Topics));
                     Context.Watch(Sender);
                     CheckOverlappingRequests("RequestMessages", Sender, requestMessages.Topics);
-                    _requests = _requests.SetItem(Sender, requestMessages);
-                    _requestors = _requestors.Add(Sender);
-                    
+
+                    if (_stageActorsMap.GetOrElse(requestMessages.Topics, Sender).Equals(Sender))
```
Aaronontheweb, line 223: If someone else already owns that exact same topic, make sure they're the same actor as the one sending the current `RequestMessages` - otherwise, ignore the request.

Aaronontheweb: If no one owns the topic yet, save the request.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -187,80 +197,96 @@ protected override bool Receive(object message)
...
+                        foreach (var offset in seek.Offsets)
+                        {
+                            _seekedOffset = _seekedOffset.SetItem(offset.TopicPartition, offset);
+                        }
+
+                        Sender.Tell(Done.Instance);
+                    }
+                    catch (Exception ex)
+                    {
+                        SendFailure(ex, Sender);
```
Aaronontheweb, line 252: Important detail: if we fail to `Seek` then we send that back as a `Status.Failure` to the `StageActor` now - this will cause that stage to fail and terminate.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -187,80 +197,96 @@ protected override bool Receive(object message)
...
+
                 case KafkaConnectionFailed kcf:
                     ProcessError(kcf);
                     Self.Tell(KafkaConsumerActorMetadata.Internal.Stop.Instance);
                     return true;
 
                 case Terminated terminated:
                     _requests = _requests.Remove(terminated.ActorRef);
-                    _requestors = _requestors.Remove(terminated.ActorRef);
+                    _stageActorsMap = _stageActorsMap.Where(c => !c.Value.Equals(terminated.ActorRef))
```
Aaronontheweb, line 274: Remove all partitions that belonged to a given actor from the `_stageActorsMap`

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -355,22 +396,24 @@ private void HandleSubscription(KafkaConsumerActorMetadata.Internal.ISubscriptio
...
 
                     case KafkaConsumerActorMetadata.Internal.AssignWithOffset assignWithOffset:
                     {
-                        var topicPartitions = assignWithOffset.TopicPartitionOffsets.Select(o => o.TopicPartition).ToImmutableHashSet();
+                        var topicPartitions = assignWithOffset.TopicPartitionOffsets.Select(o => o.TopicPartition)
+                            .ToImmutableHashSet();
                         CheckOverlappingRequests("AssignWithOffset", Sender, topicPartitions);
-                        
-                        // TODO: dear lord this is wrong, WE SHOULD NOT BE SETTING OFFSETS TO ZERO HERE
-                        var previousAssigned = _consumer.Assignment.Select(tp => new TopicPartitionOffset(tp, new Offset(0)));
```
Aaronontheweb, line 368: Fixed this code - explicitly assigning `0` here is a no-no. Instead we use whatever the most recently stored offset for this consumer on this partition was. If the consumer is using external offset storage then this should default to `Offset.Unassigned`.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -424,17 +468,28 @@ private void SchedulePollTask()
             Timers.StartSingleTimer(PollTimerKey, _pollMessage, _settings.PollInterval);
         }
 
-        private void CheckOverlappingRequests(string updateType, IActorRef fromStage, IImmutableSet<TopicPartition> topics)
+        private void RequestDelayedPoll()
```
Aaronontheweb, line 468: Encapsulated this into its own function

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -424,17 +468,28 @@ private void SchedulePollTask()
...
+                Self.Tell(_delayedPollMessage);
+            }
+        }
+
+        private void CheckOverlappingRequests(string updateType, IActorRef fromStage,
+            IImmutableSet<TopicPartition> topics)
         {
             // check if same topics/partitions have already been requested by someone else,
             // which is an indication that something is wrong, but it might be alright when assignments change.
-            foreach (var (actorRef, request) in _requests.ToTuples())
```
Aaronontheweb, line 431: `.ToTuples` is generally not needed - you can just do deconstruction on all `KeyValuePair`s coming out of an `IDictionary`

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -573,74 +631,87 @@ private void Poll()
...
                     _requests = _requests.Remove(stageActorRef);
                 }
-            }                    
+            }
+        }
+
+        private void SendFailure(Exception ex, IActorRef stageActorRef)
+        {
+            stageActorRef.Tell(new Status.Failure(ex));
+            _stageActorsMap = _stageActorsMap.Where(c => !c.Value.Equals(stageActorRef)).ToImmutableDictionary();
```
Aaronontheweb, line 655: Clears "failed" stages out of the `_stageActorMap` - those stages will also call `.Fail` on themselves in response to this `Status.Failure`

src/Akka.Streams.Kafka/Stages/Consumers/Concrete/CommittableSubSourceStage.cs
```diff
@@ -13,44 +14,85 @@
...
+                _subSourceStageLogicFactory,
+                attributes: inheritedAttributes);
             return (logic, logic.Control);
         }
         
-        /// <summary>
-        /// Creates message builder for sub-source logic
-        /// </summary>
-        private CommittableSourceMessageBuilder<K, V> GetMessageBuilder(SubSourceLogic<K, V, CommittableMessage<K, V>> logic)
+        private class
```
Aaronontheweb, line 58: Implemented the `IMessageBuilder` and all commitable-source details inside the `CommittableSubSourceStageLogicFactory`

src/Akka.Streams.Kafka/Stages/Consumers/Concrete/PlainSubSourceStage.cs
```diff
@@ -50,19 +49,46 @@ public PlainSubSourceStage(ConsumerSettings<K, V> settings, IAutoSubscription su
             Subscription = subscription;
             GetOffsetsOnAssign = getOffsetsOnAssign;
             OnRevoke = onRevoke;
+            _subSourceStageLogicFactory = new PlainSubSourceStageLogicFactory();
         }
 
-        /// <inheritdoc />
+        private class
+            PlainSubSourceStageLogicFactory : SubSourceLogic.ISubSourceStageLogicFactory<K, V, ConsumeResult<K, V>>
```
Aaronontheweb, line 56: Same idea as before - move all of the stream stage construction inside its own factory.

Aaronontheweb (commented):
Detailed my changes

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -187,80 +194,96 @@ protected override bool Receive(object message)
...
-                
+
                 case KafkaConsumerActorMetadata.Internal.Seek seek:
-                    foreach (var offset in seek.Offsets)
+                    try
                     {
-                        _seekedOffset = _seekedOffset.SetItem(offset.TopicPartition, offset);
+                        foreach (var offset in seek.Offsets)
+                        {
+                            _consumer.Seek(offset);
```
Aaronontheweb, line 245: Undid some of the changes in #225 here because the way `_seekedOffset` worked was vulnerable to a race condition that could cause the `PlainPartitionSourceWithOffset` to hang. 

I think the reason why `Seek` could fail in the first place was because the partitions that were being requested were not yet assigned to the `_consumer` requesting them. Either that or the `Seek` was called too soon after the initial assignment, ala https://github.com/confluentinc/confluent-kafka-dotnet/issues/1303 (this is my guess)

I think the changes we've made to the `KafkaConsumerActor` largely alleviate the need for this work-around - might have to run a benchmark to be certain.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
Arkatufus: The chaining needs to be moved to where `AddToPartitionAssignmentHandler` is called, the `AddToPartitionAssignmentHandler` method is supposed to be empty.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
Arkatufus: So this is basically the reverse map of `_requests` now?

Arkatufus (changes requested):
Small suggestion.

Arkatufus (approved):
LGTM

Aaronontheweb:
> So this is basically the reverse map of _requests now?

for a normal single consumer it doesn't do too much - it's more useful for partitioned streams where sub-stages are created / deallocated on-the-fly. It's designed to prevent overlapping requests from them in order to ensure that messages are not lost.

PR #453, opened 2025-03-12
Add `RebalanceExtTests` for testing partitioned message loss on rebalance

### LINKED ISSUE #447: Partition Rebalance + Data Consistency Issues

Aaronontheweb:
## Changes

Partially resolves #447 - this mostly assures that changes we've already made work as intended.

Also, expands the `ConsumerSettings` to now configure the maximum number of polled records.

## Checklist

### DIFF
--- a/src/Akka.Streams.Kafka.Tests/Integration/RebalanceExtTests.cs
+++ b/src/Akka.Streams.Kafka.Tests/Integration/RebalanceExtTests.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Akka.Streams.Dsl;
+using Akka.Streams.Kafka.Dsl;
+using Akka.Streams.Kafka.Helpers;
+using Akka.Streams.Kafka.Messages;
+using Akka.Streams.Kafka.Settings;
+using Akka.Util.Internal;
+using Confluent.Kafka;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Kafka.Tests.Integration;
+
+public class RebalanceExtTests : KafkaIntegrationTests
+{
+    public RebalanceExtTests(ITestOutputHelper output, KafkaFixture fixture)
+        : base(nameof(RebalanceExtTests), output, fixture)
+    {
+    }
+
+    private const string ConsumerClientId1 = "consumer-1";
+    private const string ConsumerClientId2 = "consumer-2";
+
+    private sealed record TopicPartitionMetadata(
+        IReadOnlyCollection<string> Topics,
+        IReadOnlyCollection<TopicPartition> Partitions,
+        IReadOnlyDictionary<string, TaskCompletionSource<Done>> TpFutureMap,
+        IReadOnlyCollection<Task<Done>> ProducerTpsAck,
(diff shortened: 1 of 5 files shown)


src/Akka.Streams.Kafka/Settings/ConsumerSettings.cs
```diff
@@ -160,6 +161,14 @@ public static ConsumerSettings<TKey, TValue> Create(Akka.Configuration.Config co
...
         /// </remarks>
         public bool VerboseLogging { get; init; } = false;
+
+        /// <summary>
+        /// Maximum number of records to fetch per poll attempt.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to 10.
+        /// </remarks>
+        public int MaxPollRecords { get; init; } = 10;
```
Aaronontheweb, line 171: Allow the end-user to specify the maximum number of records polled as a tuning parameter - this is something you can set directly inside the Java Kafka client but not .NET's due to some design differences, so we're kind of simulating that here.

src/Akka.Streams.Kafka.Tests/Integration/RebalanceExtTests.cs
```diff
@@ -0,0 +1,381 @@
...
+                    return Done.Instance;
+                }
+            }
+        }
+
+        return (producerTpsAck, messageAndStoreAck);
+    }
+
+    [Fact(DisplayName =
+        "Fetched records must not be lost when two consumers consume from one topic and two partitions and one consumer aborts mid-stream")]
```
Aaronontheweb, line 213: This is a very complicated spec that I ported from https://github.com/akka/alpakka-kafka/releases/tag/v3.0.1 - it doesn't translate 1:1 with .NET because we don't have fine-grained control over partition assignments, but it does assert that we don't lose messages even if one partitioned consumer gets aborted mid-flight.

src/Akka.Streams.Kafka/reference.conf
```diff
@@ -33,6 +33,10 @@ akka.kafka.consumer {
   # is executing the stage will be blocked. See also the `wakup-timeout` setting below.
   poll-timeout = 50ms
   
+  # Tuning property that indicates how many records to poll from the Kafka consumer
+  # in any given attempt.
+  max-poll-records = 10
```
Aaronontheweb, line 38: Added equivalent HOCON setting

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -581,7 +581,7 @@ private void Poll()
 
         private (List<ConsumeResult<K, V>>, Exception?) PollKafka(CancellationToken token)
         {
-            var i = 10; // 10 poll attempts
+            var i = _settings.MaxPollRecords; // use the number of poll attempts specified in the settings
```
Aaronontheweb, line 584: Use the new setting we defined inside the `KafkaConsumerActor`

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #458, opened 2025-03-14
Adding support for configurable commit collecting / draining

### LINKED ISSUE #447: Partition Rebalance + Data Consistency Issues

Aaronontheweb:
## Changes

Final piece of https://github.com/akkadotnet/Akka.Streams.Kafka/issues/447

close #447

## Checklist

* [x] Design discussion issue #447


src/Akka.Streams.Kafka.Tests/BugFix240SupervisionStrategy.cs
```diff
@@ -215,7 +215,9 @@ await Source.From(Enumerable.Range(1, 11))
                 })
                 .SelectAsync(1, async elem =>
                 {
+#pragma warning disable CS0618 // Type or member is obsolete
                     await elem.CommitableOffset.Commit();
```
Aaronontheweb, line 219: Fixing this is going to require more public API changes

src/Akka.Streams.Kafka.Tests/Internal/CommitCollectorStageSpecs.cs
Aaronontheweb, line 1: Big test suite for covering the `CommitCollectorStage` and its `CommitWhen` modes.

src/Akka.Streams.Kafka/Helpers/Committer.cs
```diff
@@ -17,12 +18,11 @@ public static class Committer
...
         {
-            return Akka.Streams.Dsl.Flow.Create<ICommittable>().GroupedWithin(settings.MaxBatch, settings.MaxInterval)
-                .Select(CommittableOffsetBatch.Create)
-                .SelectAsync(settings.Parallelism, async batch =>
+            return Akka.Streams.Dsl.Flow.FromGraph(new CommitCollectorStage(settings))
+                .SelectAsyncUnordered(settings.Parallelism, async batch =>
                 {
-                   await batch.Commit();
-                   return batch;
+                    await ((CommittableOffsetBatch)batch).Commit();
```
Aaronontheweb, line 24: We should probably consider either doing something with the interfaces here or moving this to a base class so I can safely call an `internal` Commit method.

src/Akka.Streams.Kafka/Helpers/Committer.cs
```diff
@@ -17,12 +18,11 @@ public static class Committer
         /// </summary>
         public static Flow<ICommittable, ICommittableOffsetBatch, NotUsed> BatchFlow(CommitterSettings settings)
         {
-            return Akka.Streams.Dsl.Flow.Create<ICommittable>().GroupedWithin(settings.MaxBatch, settings.MaxInterval)
-                .Select(CommittableOffsetBatch.Create)
-                .SelectAsync(settings.Parallelism, async batch =>
+            return Akka.Streams.Dsl.Flow.FromGraph(new CommitCollectorStage(settings))
```
Aaronontheweb, line 21: Uses the new `CommitCollectorStage`

src/Akka.Streams.Kafka/Messages/Committable.cs
```diff
@@ -13,28 +14,36 @@ public interface ICommittable
...
+        
         /// <summary>
         /// Get a number of processed messages this committable contains
         /// </summary>
         long BatchSize { get; }
+        
+        /// <summary>
+        /// Add/overwrite an offset position from another committable.
+        /// </summary>
+        ICommittableOffsetBatch Updated(ICommittable offset);
```
Aaronontheweb, line 28: Added this so we can easily convert from an individual commit to a batch.

src/Akka.Streams.Kafka/Messages/Committable.cs
```diff
@@ -13,28 +14,36 @@ public interface ICommittable
         /// <summary>
         /// Commits an offset that is included in a <see cref="CommittableMessage{K,V}"/> 
         /// </summary>
+        [Obsolete("Do not use directly. Use Committer.Flow or Committer.Sink instead.")]
```
Aaronontheweb, line 17: We want to discourage users from doing a `SelectAsync` + `await .Commit()` because that means losing all of the data-loss protection that's built into the `CommitCollectStage`. We can still use this API internally but users should really use the `Commiter.Sink` or `Committer.Flow` <- those aren't even the correct names for those stages right now due to some really poor API design decisions but that's what they'll eventually be called

src/Akka.Streams.Kafka/Messages/CommittableOffset.cs
```diff
@@ -10,22 +9,29 @@ namespace Akka.Streams.Kafka.Messages
...
-        /// Cosumed record metadata
+        /// Consumed record metadata
         /// </summary>
         public string Metadata { get; }
+        
         /// <summary>
         /// Committer
         /// </summary>
-        public IInternalCommitter Committer { get; }
+        public KafkaAsyncConsumerCommitter Committer { get; }
```
Aaronontheweb, line 32: `IInternalCommitter` wasn't a super useful abstraction and I removed it - making the parts of the `KafkaAsyncConsumerCommitter` `virtual` where needed instead.

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
```diff
@@ -17,7 +18,7 @@ internal sealed class CommittableOffsetBatch : ICommittableOffsetBatch
         /// CommittableOffsetBatch
         /// </summary>
         public CommittableOffsetBatch(IImmutableDictionary<GroupTopicPartition, OffsetAndMetadata> offsetsAndMetadata, 
-                                      IImmutableDictionary<string, IInternalCommitter> committers, 
+                                      IImmutableDictionary<GroupTopicPartition, KafkaAsyncConsumerCommitter> committers,
```
Aaronontheweb, line 21: We had a blend of `string`s and other stuff flying around. Re-organized this on `GroupTopicPartition` for `committers` since those _do_ need to incorporate `groupId` and `Partition` in addition to topic.

src/Akka.Streams.Kafka/Messages/CommittableOffsetBatch.cs
```diff
@@ -86,66 +85,82 @@ public ICommittableOffsetBatch Updated(ICommittable offset)
...
+        internal CommittableOffsetBatch Filter(Predicate<GroupTopicPartition> p)
+        {
+            var newOffsets = OffsetsAndMetadata.Where(o => p(o.Key))
+                .ToImmutableDictionary(o => o.Key, o => o.Value);
+            var newCommiters =
+                Offsets.ToImmutableDictionary(c => c.GroupTopicPartition, v => CommitterFor(v.GroupTopicPartition));
+            return new CommittableOffsetBatch(newOffsets, newCommiters, BatchSize);
+        }
+        
+        public CommittableOffsetBatch TellCommit()
```
Aaronontheweb, line 148: These methods all produce the commit messages that we're going to send to the `KafkaConsumerActor`.

src/Akka.Streams.Kafka/Messages/GroupTopicPartitionOffset.cs
```diff
@@ -43,111 +43,20 @@ public GroupTopicPartitionOffset(GroupTopicPartition groupTopicPartition, Offset
...
-        {
-            unchecked
-            {
-                return (Offset.GetHashCode() * 397) ^ (Metadata != null ? Metadata.GetHashCode() : 0);
-            }
-        }
+        public TopicPartition TopicPartition { get; } = new(Topic, Partition);
     }
+    
+    public sealed record OffsetAndMetadata(Offset Offset, string Metadata);
```
Aaronontheweb, line 61: We should really remove `OffsetAndMetadata` because we don't even really use `Metadata` at all - this is something that got ported from the JVM without understanding what it meant. This API wasn't an Akka thing, it was a Confluent.Kafka for Java thing. 

In the .NET client we just use `TopicPartitionOffset` instead - so deleting this in the long run would be a good idea / would reduce waste. I tried doing this on this PR and it just ballooned the size of an already large PR even larger, so better to make this its own project.

src/Akka.Streams.Kafka/Settings/CommitterSettings.cs
```diff
@@ -1,25 +1,65 @@
...
+using Akka.Streams.Kafka.Messages;
 
 namespace Akka.Streams.Kafka.Settings
 {
+    /// <summary>
+    /// Determines when to commit offsets
+    /// </summary>
+    public interface ICommitWhen;
+
+    public static class CommitWhen
```
Aaronontheweb, line 13: `ICommitWhen` strategy - specifies when we want to commit:

1. `OffsetFirstObserved` - commit observations as we see them
2. `NextOffsetObserved` - when commit A arrives, don't commit it until B arrives.

The reason for this is that `NextOffsetObserved` is a little bit tighter from an "at-least-once-delivery" standpoint, but it does run the higher risk of generating a duplicate - whereas `OffsetFirstObserved` commits more aggressively.

src/Akka.Streams.Kafka/Settings/CommitterSettings.cs
```diff
@@ -1,25 +1,65 @@
...
+                    $"Allow values are offset-first-observed, next-offset-observed.Received: {s}")
+            };
+        }
+    }
+    
     /// <summary>
     /// Settings for committer. See 'akka.kafka.committer' section in reference.conf.
     /// </summary>
-    public sealed class CommitterSettings
+    public sealed record CommitterSettings
```
Aaronontheweb, line 48: Turning stuff into `record`s as part of the public API to cut down on the amount of copying / equality stuff going on.

src/Akka.Streams.Kafka/Settings/CommitterSettings.cs
```diff
@@ -1,25 +1,65 @@
...
+        /// <summary>
+        /// Commit the previous offset as soon as a new <see cref="ICommittableOffset"/> is observed.
+        /// </summary>
+        public sealed class NextOffsetObserved : ICommitWhen
+        {
+            public static NextOffsetObserved Instance { get; } = new NextOffsetObserved();
+            private NextOffsetObserved() { }
+        }
+
+        public static ICommitWhen ValueOf(string s)
```
Aaronontheweb, line 33: Method for parsing `ICommitWhen` from HOCON.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -75,16 +77,15 @@ internal class KafkaConsumerActor<K, V> : ActorBase, ILogReceive, IWithTimers
         private readonly Decider _decider;
 
         /// <summary>
-        /// While `true`, committing is delayed.
-        /// Changed by `onPartitionsRevoked` and `onPartitionsAssigned` callbacks
+        /// Collect commit offset maps until the next poll
         /// </summary>
-        private bool _rebalanceInProgress = false;
-
+        private IImmutableList<TopicPartitionOffset> _commitMaps =
```
Aaronontheweb, line 82: Set of all commits that are going to need to be aggregated and committed.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -75,16 +77,15 @@ internal class KafkaConsumerActor<K, V> : ActorBase, ILogReceive, IWithTimers
         private readonly Decider _decider;
 
         /// <summary>
-        /// While `true`, committing is delayed.
-        /// Changed by `onPartitionsRevoked` and `onPartitionsAssigned` callbacks
+        /// Collect commit offset maps until the next poll
         /// </summary>
-        private bool _rebalanceInProgress = false;
```
Aaronontheweb, line 81: Don't need this anymore.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -75,16 +77,15 @@ internal class KafkaConsumerActor<K, V> : ActorBase, ILogReceive, IWithTimers
...
+        private IImmutableList<TopicPartitionOffset> _commitMaps =
+            ImmutableList<TopicPartitionOffset>.Empty;
+        
         /// <summary>
-        /// Keeps commit offsets during rebalances for later commit.
+        /// Keep commit senders that need a reply once stashed commits are made
         /// </summary>
-        private IImmutableSet<TopicPartitionOffset>
-            _rebalanceCommitStash = ImmutableHashSet<TopicPartitionOffset>.Empty;
+        private ImmutableHashSet<IActorRef> _commitSenders = ImmutableHashSet<IActorRef>.Empty;
```
Aaronontheweb, line 88: Keep track of which actors have sent us pending commits that need to be ACK-ed.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -230,19 +223,34 @@ protected override bool Receive(object message)
         {
             switch (message)
             {
-                case KafkaConsumerActorMetadata.Internal.Commit commit when _rebalanceInProgress:
-                    _rebalanceCommitStash = _rebalanceCommitStash.Union(commit.Offsets);
-                    _rebalanceCommitSenders = _rebalanceCommitSenders.Add(Sender);
-                    return true;
-
                 case KafkaConsumerActorMetadata.Internal.Commit commit:
-                    _commitRefreshing.Add(commit.Offsets);
```
Aaronontheweb, line 239: Changed what `Commit` does, as it was basically a bulk commit-everything-now message before. Now it gives us new `TopicPartitionOffset`s to add to the `_commitMaps` - and we will commit this stuff in aggregate in the near future.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -230,19 +223,34 @@ protected override bool Receive(object message)
...
+                    // prepending as later received offsets are most likely higher
+                    _commitMaps = ImmutableList<TopicPartitionOffset>.Empty
+                        .Add(commit.TopicPartitionOffset)
+                        .AddRange(_commitMaps);
+                    _commitSenders = _commitSenders.Add(Sender);
                     return true;
 
-                case Internal.Poll<K, V> poll:
-                    ReceivePoll(poll);
+                case KafkaConsumerActorMetadata.Internal.CommitWithoutReply commitWithoutReply:
```
Aaronontheweb, line 234: `CommitWithoutReply` is fire-and-forget committing - usually happens during an emergency commit when things are shutting down.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -230,19 +223,34 @@ protected override bool Receive(object message)
...
+                    // prepending as later received offsets are most likely higher
+                    _commitMaps = ImmutableList<TopicPartitionOffset>.Empty
+                        .Add(commitWithoutReply.TopicPartitionOffset)
+                        .AddRange(_commitMaps);
+
+                    if (commitWithoutReply.Emergency)
+                    {
+                        EmergencyPoll();
```
Aaronontheweb, line 242: if this was indeed an emergency - we will commit and poll _right now_ rather than waiting until later. Happens during stage completions and failures when there are pending commits that have not been confirmed.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -230,19 +223,34 @@ protected override bool Receive(object message)
...
+                    }
+                    
+                    return true;
+                
+                case KafkaConsumerActorMetadata.Internal.CommitSingle commitSingle:
```
Aaronontheweb, line 247: `CommitSingle` for doing commits in non-batch fashion.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -551,15 +573,7 @@ private void ReceivePoll(Internal.Poll<K, V> poll)
             // We overloaded `==`, we need to use `ReferenceEquals` to do this
             if (ReferenceEquals(poll.Target, this))
             {
-                var refreshOffsets = _commitRefreshing.RefreshOffsets;
```
Aaronontheweb, line 554: Moved all of this stuff to `CommitAndPoll`

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -572,11 +586,57 @@ private void ReceivePoll(Internal.Poll<K, V> poll)
             }
         }
 
+        private void CommitAggregatedOffsets()
+        {
+            if(_commitMaps.Count == 0) return;
+            var aggregatedOffsets = AggregateOffsets(_commitMaps);
+            // commits can occur after the partition has been revoked from the consumer, so ensure that we only attempt to
```
Aaronontheweb, line 593: This comment details a big part of how we do rebalance handling - we filter failed / will-fail-on-next-attempt commits out here after those partitions are no longer assigned to us.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -572,11 +586,57 @@ private void ReceivePoll(Internal.Poll<K, V> poll)
...
+            var assignedOffsetsToCommit = aggregatedOffsets.Where(kvp => _consumer.Assignment.Contains(kvp.Key))
+                .Select(c => new TopicPartitionOffset(c.Key, c.Value)).ToImmutableSet();
+            var replyTo = _commitSenders;
+            // flush the data before calling `consumer.commit`
+            _commitMaps = ImmutableList<TopicPartitionOffset>.Empty;
+            _commitSenders = ImmutableHashSet<IActorRef>.Empty;
+            Commit(assignedOffsetsToCommit, replyTo);
+        }
+
+        public static IReadOnlyDictionary<TopicPartition,Offset> AggregateOffsets(IReadOnlyCollection<TopicPartitionOffset> offsets)
```
Aaronontheweb: Take all of the commits we've been sent and sort each partition to its highest offset - then commit those.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -572,11 +586,57 @@ private void ReceivePoll(Internal.Poll<K, V> poll)
...
+            return aggregate;
+        }
+        
+        private void CommitAndPoll()
+        {
+            var refreshOffsets = _commitRefreshing.RefreshOffsets;
+            if (refreshOffsets.Any())
+            {
+                _log.Debug("Refreshing committed offsets: {0}", refreshOffsets.JoinToString(", "));
+                Commit(refreshOffsets, ImmutableHashSet<IActorRef>.Empty);
```
Aaronontheweb, line 632: If any commits need to be refreshed first, we do that here.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -572,11 +586,57 @@ private void ReceivePoll(Internal.Poll<K, V> poll)
...
+            }
+            Poll();
+        }
+
         private void Poll()
         {
             var currentAssignment = _consumer.Assignment.ToImmutableList();
-            var initialRebalanceInProcess = _rebalanceInProgress;
-
+            CommitAggregatedOffsets();
```
Aaronontheweb, line 641: Aggregated offsets get committed next as this will help drive some of the polling behavior (i.e. what messages are available beyond our currently committed offsets for each partition?)

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -755,59 +818,87 @@ private void ProcessError(Exception error)
...
+        private void Commit(IImmutableSet<TopicPartitionOffset> commitMap, IImmutableSet<IActorRef> replyTo)
         {
+            var watch = Stopwatch.StartNew();
             try
             {
                 _commitRefreshing.UpdateRefreshDeadlines(commitMap.Select(tp => tp.TopicPartition)
                     .ToImmutableHashSet());
-
-                var watch = Stopwatch.StartNew();
+                _commitsInProgress += 1;
```
Aaronontheweb, line 830: `_commitsInProgress` for the JVM made more sense since their commits use `Task`s - ours are all synchronous so this probably will never be any value other than `1` or `0`.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -755,59 +818,87 @@ private void ProcessError(Exception error)
...
+                {
+                    s.Tell(Done.Instance);
+                }
             }
-            catch (Exception ex)
+            catch (TopicPartitionOffsetException offsetException)
             {
-                sendReply(new Status.Failure(ex));
+                watch.Stop();
+                if (offsetException.Error.Code == ErrorCode.RebalanceInProgress)
```
Aaronontheweb: If we had failed commits due to a `ErrorCode.RebalanceInProgress` we can retry those and they will succeed so long as the partitions they belong to weren't revoked from us. If they were revoked they'll be filtered out inside the partition assignment handler.

Aaronontheweb: I thought about inspecting each of the partition offsets for errors individually for this purpose too but decided to just send them all instead.

src/Akka.Streams.Kafka/Stages/Consumers/Actors/KafkaConsumerActor.cs
```diff
@@ -755,59 +818,87 @@ private void ProcessError(Exception error)
...
+                _commitMaps = commitMap.ToImmutableList().AddRange(_commitMaps);
+                _commitSenders = _commitSenders.Union(replyTo);
                 RequestDelayedPoll();
             }
-        }
 
-        /// <summary>
-        /// Detects state changes of <see cref="_rebalanceInProgress"/> and takes action on it.
-        /// </summary>
-        private void CheckRebalanceState(bool initialRebalanceInProgress)
```
Aaronontheweb, line 798: Don't need this method any longer.

src/Akka.Streams.Kafka/Stages/Consumers/CommitCollectorStage.cs
```diff
@@ -0,0 +1,167 @@
...
+            InheritedAttributes = inheritedAttributes;
+            ObservationLogic = new CommitObservationLogic(stage._settings);
+            SetHandler(Stage.Out, OnPull);
+            SetHandler(Stage.In, OnPush, OnUpstreamFinish, OnUpstreamFailure);
+        }
+
+        private void OnUpstreamFailure(Exception ex)
+        {
+            Log.Debug(ex, "OnUpstreamFailure with exception with {0}", ObservationLogic.OffsetBatch);
+            if (ActiveBatchInProgress)
```
Aaronontheweb, line 48: If we have active data and the stages above us fail, deliver our pending batch in the form of an emergency commit to the `KafkaConsumerActor`. We have tests that verify that this works.

src/Akka.Streams.Kafka/Stages/Consumers/CommitCollectorStage.cs
```diff
@@ -0,0 +1,167 @@
...
+            FailStage(ex);
+        }
+
+        private void OnUpstreamFinish()
+        {
+            if (ActiveBatchInProgress)
+            {
+                Log.Debug("PushDownStream triggered by {0}, outstanding batch {1}", CommitTrigger.UpstreamFinish.Instance,
+                    ObservationLogic.OffsetBatch);
+                Emit(Stage.Out, ObservationLogic.OffsetBatch);
```
Aaronontheweb, line 63: Emit the batch out, gracefully, downstream if the upstream terminates.

src/Akka.Streams.Kafka/Stages/Consumers/CommitCollectorStage.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            CompleteStage();
+        }
+
+        private void OnPush()
+        {
+            var offset = Grab(Stage.In);
+            Log.Debug("Consuming offset {0}", offset);
+            if (ObservationLogic.UpdateBatch(offset))
```
Aaronontheweb, line 72: If the latest offset we've received triggers an update to our batch (newer `Offset`) then we'll check our commit triggers and update the batch accordingly.

src/Akka.Streams.Kafka/Stages/Consumers/CommitCollectorStage.cs
```diff
@@ -0,0 +1,167 @@
...
+
+                    break;
+                }
+                default:
+                    Log.Warning("Unexpected timer [{0}]", timerKey);
+                    break;
+            }
+        }
+
+        private void PushDownStream(CommitTrigger.ITriggeredBy triggeredBy)
```
Aaronontheweb, line 153: If we hit this method it's because we were triggered by batch size or by time delay - that gets logged here but otherwise we prepare the next batch and push the current one downstream.

src/Akka.Streams.Kafka/Stages/Consumers/CommitObservationLogic.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Akka.Streams.Kafka.Stages.Consumers;
+
+/// <summary>
+/// Shared commit observation logic for different graph stage logics that facilitate offset commits.
+/// </summary>
+/// <remarks>
+/// Really should be implemented as an interface with default implementations, but we can't have nice
+/// things in .NET Standard 2.0. Therefore this gets injected a dependency instead.
+/// </remarks>
+internal sealed class CommitObservationLogic
```
Aaronontheweb, line 18: Shared logic for committing - used by the `CommitCollectorStage` and also needs to be incorporated into the transactional consumers too.

src/Akka.Streams.Kafka/reference.conf
```diff
@@ -136,4 +136,10 @@ akka.kafka.committer {
     
     # Set parallelism for async committing
     parallelism = 1
+    
+    # strategy to use when committing offsets
+    # possible values: "offset-first-observed" or "next-offset-observed"
+    # offset-first-observed: When the offset of a message has been successfully produced.
+    # next-offset-observed: When the next offset is observed.
+    when = "offset-first-observed"
```
Aaronontheweb, line 144: Default strategy is to commit as soon as new offsets are observed.

Aaronontheweb (commented):
Detailed my changes - these were ported from https://github.com/akka/alpakka-kafka/releases/tag/v3.0.1 which is all under Apache 2.0 still.

Aaronontheweb:
In addition to making committing more robust, this should significantly improve performance for things like `CommitablePartitionedSource`

Arkatufus (approved):
LGTM, we need to communicate these changes to the customers later on

Aaronontheweb:
> LGTM, we need to communicate these changes to the customers later on

I'll make sure we do a bit documentation update before we do a release

PR #454, opened 2025-03-12
Resolve `ArgumentException` upon partition re-assignment

Aaronontheweb:
## Changes

close #415 - this is fairly difficult to test in our test suite without having manual control over partition assignments, so the test lab might have better luck. I'll see what I can come up with. But in the meantime, check out my PR comments.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #415, #447

PR #464, opened 2025-03-16
fix `IControl` and ensure clean draining / termination of Akka.Streams consumers

Aaronontheweb:
## Changes

Noticed while working on https://github.com/akkadotnet/Akka.Streams.Kafka/pull/458 and https://github.com/akkadotnet/Akka.Streams.Kafka/pull/454 that the `IControl.IsShutdown` never seems to complete even when the stream stages shut down, which is 100% a bug.

## Checklist

* [ ] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).


src/Akka.Streams.Kafka/Helpers/Control.cs
```diff
@@ -23,7 +22,7 @@ public interface IControl
         /// Shutdown the consumer `Source`. It will wait for outstanding offset
         /// commit requests to finish before shutting down.
         /// </summary>
-        Task Shutdown(Exception? ex = null);
```
Aaronontheweb, line 26: This method should not take an `Exception` and, in fact, it was often our Exception-handling inside the `IControl` implementations that was causing them to hang and not work. Breaking API change but given that this was unusable before I think it's probably NBD.

src/Akka.Streams.Kafka/Helpers/Control.cs
```diff
@@ -41,95 +40,141 @@ public interface IControl
         Task<TResult> DrainAndShutdown<TResult>(Task<TResult> streamCompletion);
     }
 
+    /// <summary>
+    /// Helper class for creating a <see cref="DrainingControl{T}"/> instances.
+    /// </summary>
+    public static class DrainingControl
+    {
+        public static DrainingControl<T> Create<T>(IControl control, Task<T> streamCompletion) =>
```
Aaronontheweb, line 47: Developer Experience: it's much easier to invoke `static Class.Method<T>` as a method because the `T` parameter can be inferred, whereas with `static Class<T>.Method` you have to know in advance what your `T` is. I added this to support the former.

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,86 @@
 using System;
 using System.Threading.Tasks;
 using Akka.Streams.Kafka.Extensions;
-using Akka.Streams.Util;
 
-#nullable enable
 namespace Akka.Streams.Kafka.Helpers
 {
+    internal static class PromiseControl
```
Aaronontheweb, line 6: Instead of passing in `Exception`s to signal what type of `AsyncCallback` we have to invoke, we use the `IControlOperation` types and pattern-match instead.

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,86 @@
...
+            private ControlShutdown(){}
+        }
+    }
+    
     /// <summary>
     /// Used in source logic classes to provide <see cref="IControl"/> implementation.
     /// </summary>
     /// <typeparam name="TSourceOut"></typeparam>
-    internal abstract class PromiseControl<TSourceOut> : IControl
+    internal class PromiseControl<TSourceOut> : IControl
```
Aaronontheweb, line 27: There are some sub-classes for `PromiseControl<T>` but honestly we could probably make this a `sealed` class if we really wanted to - but I'm cool with it as-is since it's all `internal` anyway.

src/Akka.Streams.Kafka/Helpers/Control.cs
```diff
@@ -41,95 +39,143 @@ public interface IControl
...
+            DrainingControl<T>.Create(control, streamCompletion);
+
+        /// <summary>
+        /// Stop producing messages from the `Source`, wait for stream completion
+        /// and shut down the consumer `Source` so that all consumed messages
+        /// reach the end of the stream.
+        /// Failures in stream completion will be propagated, the source will be shut down anyway.
+        /// </summary>
+        internal static async Task<TResult> DrainAndShutdownDefaultAsync<TResult>(this IControl control,
```
Aaronontheweb, line 56: **BREAKING API CHANGE**

This method never should have been exposed directly to end users - so I removed it from the `public` API and made it part of the `DrainingControl` class.

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,85 @@
 using System;
 using System.Threading.Tasks;
-using Akka.Streams.Kafka.Extensions;
-using Akka.Streams.Util;
 
-#nullable enable
```
Aaronontheweb, line 6: Not needed now that nullability is enabled solution-wide.

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,85 @@
...
         private readonly Action<Outlet<TSourceOut>> _completeStageOutlet;
         private readonly Action<bool> _setStageKeepGoing;
 
-        private readonly TaskCompletionSource<Done> _shutdownTaskSource = new TaskCompletionSource<Done>();
-        private readonly TaskCompletionSource<Done> _stopTaskSource = new TaskCompletionSource<Done>();
-        private readonly Action _stopCallback;
-        private readonly Action<Exception?> _shutdownCallback;
+        private readonly TaskCompletionSource<Done> _shutdownTaskSource = new();
+        private readonly TaskCompletionSource<Done> _stopTaskSource = new();
+        private readonly Action<PromiseControl.IControlOperation> _controlCallback;
```
Aaronontheweb, line 35: We take an `IControlOperation` instead of an `Exception` now.

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,85 @@
...
+        private readonly Action _performShutdown;
```
Aaronontheweb, line 36: Shutdown invocation

src/Akka.Streams.Kafka/Helpers/PromiseControl.cs
```diff
@@ -1,65 +1,85 @@
...
             _shape = shape;
             _completeStageOutlet = completeStageOutlet;
             _setStageKeepGoing = setStageKeepGoing;
-
-            _stopCallback = asyncCallbackFactory(PerformStop);
-            _shutdownCallback = asyncShutdownCallbackFactory(PerformShutdown);
+            _performShutdown = performShutdown;
+            _controlCallback = asyncShutdownCallbackFactory(c =>
+            {
+                switch (c)
```
Aaronontheweb, line 51: This is one of three big fixes: these mechanics basically force the correct shutdown method to be invoked properly, which wasn't happening before.

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -69,14 +70,14 @@ protected BaseSingleSourceLogic(
             _shape = shape;
             Subscription = subscription;
             _messageBuilder = messageBuilderFactory(this);
-            Control = new BaseSingleSourceControl(_shape, Complete, SetKeepGoing, GetAsyncCallback, GetAsyncCallback,
+            Control = new PromiseControl<TMessage>(_shape, Complete, SetKeepGoing, GetAsyncCallback,
```
Aaronontheweb, line 73: No need for a child class here

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/BaseSingleSourceLogic.cs
```diff
@@ -232,9 +235,8 @@ protected virtual void MessageHandling((IActorRef, object) args)
 
                     break;
 
-                case Terminated terminated:
-                    if (Log.IsInfoEnabled)
-                        Log.Info("Consumer actor terminated: {0}", terminated.ActorRef.Path);
```
Aaronontheweb, line 237: Second big fix: this is probably the source of several supervision bugs, but the death pact between the stream stage and its consumer actor needs to be honored

Aaronontheweb: This is probably where a bunch of our supervision errors came from

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -533,29 +528,22 @@ private void PerformShutdown(Exception? ex)
...
-                _performShutdown = performShutdown;
             }
-
-            /// <inheritdoc />
-            public override void PerformStop() => _performStop();
-
-            /// <inheritdoc />
-            public override void PerformShutdown(Exception? ex) => _performShutdown(ex);
+            
+            protected override void PerformStop() => _performStop();
```
Aaronontheweb, line 546: Uses a custom stopping function that isn't needed for single sources

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -720,8 +699,8 @@ private StageActorRef.Receive MessageHandling() => args =>
 
         public override void PostStop()
         {
+            // TODO: need to try to shut consumer actor down here
```
Aaronontheweb, line 702: Third bug: if we shut down, we need to make sure we kill off the `KafkaConusmerActor` here - this will help the `IControl` complete shutdown too.

I'll do another PR for this

src/Akka.Streams.Kafka/Extensions/ControlExtensions.cs
```diff
@@ -7,39 +7,8 @@ namespace Akka.Streams.Kafka.Extensions
     /// <summary>
     /// ControlExtensions
     /// </summary>
-    public static class ControlExtensions
+    internal static class ControlExtensions
```
Aaronontheweb, line 10: **BREAKING API CHANGE**

Moved this to `internal` and gutted this class. These implementation details should have been internal all along - and generally, the namespaces this plugin uses are in dire need of massive re-organization (namely, there should be fewer of them.)

Aaronontheweb (commented):
Detailed my changes, including some breaking changes I made. IMHO we should probably break and reset the bone here by paring down the namespaces and re-evaluating whether some things should be better encapsulated or not. We can always SemVer it if the need was dire but then we'd no longer be versioning along Akka.NET's lines.

I'm persuadable either way and I can "unbreak" the changes by marking things as `Obsolete` instead.

src/Akka.Streams.Kafka.Tests/Integration/RebalanceExtTests.cs
```diff
@@ -350,8 +350,7 @@ The alternative to doing this is re-structuring the test to limit event producti
             await Task.WhenAll(topicMetadata.TpFutureMap.Values.Select(t => t.Task)).WaitAsync(RemainingOrDefault);
 
             // shutdown the consumers
-            Assert.True(control1.Shutdown().IsCompleted);
-            //Assert.True(control2.Shutdown().IsCompleted); // TODO: this is still buggy
+            await Task.WhenAll(control1.Shutdown(), control2.Shutdown()).WaitAsync(RemainingOrDefault);
```
Aaronontheweb, line 353: FYI this assertion would fail without these fixes

src/Akka.Streams.Kafka.Tests/Integration/RebalanceIntegrationTests.cs
Aaronontheweb, line 1: The `IControl` assertions in this class would fail without these fixes

src/Akka.Streams.Kafka.Tests/Dsl/ControlSpec.cs
```diff
@@ -0,0 +1,81 @@
+using System;
+using System.Threading.Tasks;
+using Akka.Streams.Kafka.Helpers;
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Streams.Kafka.Tests.Dsl;
+
+public class ControlSpec
```
Aaronontheweb, line 9: Verifies that the `IControl` and `DrainingControl` `Task` composition work and make sense.

src/Akka.Streams.Kafka/Helpers/Control.cs
```diff
@@ -41,95 +39,143 @@ public interface IControl
...
+            Task<TResult> streamCompletion)
+        {
+            TResult result;
+
+            try
```
Aaronontheweb, line 61: Had to re-structure the `Task` and `await` composition significantly here in order to get `ControlSpec` to pass.

Aaronontheweb (commented):
Added some more details

Arkatufus (approved):
LGTM, we'll need to do consumer acceptance for all of the API changes though, most probably by doing a beta release

PR #468, opened 2025-03-18
Migrated `ConnectionCheckerSettings` to use `record` types

Aaronontheweb:
## Changes

Get rid of old `.Copy` methods where we can.

## Checklist


Aaronontheweb:
Changes in API should all be source-compatible.

PR #472, opened 2025-03-19
Performance tuning consumer and committer settings

Aaronontheweb:
## Changes

Set `akka.kafka.consumer.max-poll-records` to 500, which is the Kafka default: https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#max-poll-records

Set `akka.kafka.committer.parallelism = 100` - was 1 previously

## Checklist


src/Akka.Streams.Kafka/reference.conf
```diff
@@ -35,7 +35,7 @@ akka.kafka.consumer {
   
   # Tuning property that indicates how many records to poll from the Kafka consumer
   # in any given attempt.
-  max-poll-records = 10
+  max-poll-records = 500
```
Aaronontheweb, line 38: Set this to 500, which has consistently yielded the best results from our benchmarks (because it's equal to the internal poll max-buffer-size)

src/Akka.Streams.Kafka/reference.conf
```diff
@@ -135,7 +136,7 @@ akka.kafka.committer {
     max-interval = 10s
     
     # Set parallelism for async committing
-    parallelism = 1
+    parallelism = 100
```
Aaronontheweb, line 139: Set the parallelism for committing equal to 100 in-flight commit attempts at a time. These will get aggregated together by the `KafkaConsumerActor` and consolidated to highest-offset-per-partition, so there's no consistency issues there. This just allows the stream to demand more while commits are in-process.

Aaronontheweb (commented):
Detailed my config changes

Arkatufus (approved):
LGTM

PR #492, opened 2025-05-29
Refactor simple example as Aspire project

Arkatufus:


Arkatufus:
Some problems encountered while implementing this:
* Akka.Streams.Kafka.Cpu.Benchmark project would suddenly fail to resolve its dependency
   * A `dotnet restore --force` usually fixes this
* Aspire Build and run failed with error message "Unable to get the project output for net8.0" when the target framework is clearly net9.0
   * Clean the project
      * Delete all of the `bin` and `obj` folders in the solution folder
      * Do `dotnet clean`
      * Do `dotnet restore --force`
   * If that fails, try checking the dotnet workload, try one of these inside the solution folder, if that doesn't work, try them inside the aspire AppHost project folder:
      * `dotnet workload update`
      * `dotnet workload install aspire`
      * `dotnet workload uninstall aspire`
   * Note that the last command actually uninstall the aspire workload. The net9.0 aspire is supposed to not rely on workloads anymore, its supposed to set everything up from the SDK nuget package.
   * If all else fails, try restarting your IDE

examples/Simple/SimpleConsumer.AppHost/SimpleConsumer.AppHost.csproj
```diff
@@ -0,0 +1,24 @@
...
+
+  <Sdk Name="Aspire.AppHost.Sdk" Version="$(AspireVersion)"/>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>$(NetCoreTestVersion)</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <UserSecretsId>df5f361f-692c-4438-a33f-9f6b45342370</UserSecretsId>
```
Aaronontheweb: we can probably remove this

examples/Simple/SimpleConsumer.AppHost/Stub.cs
```diff
@@ -0,0 +1,7 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="Stub.cs" company="Akka.NET Project">
+//      Copyright (C) 2025 - 2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+// -----------------------------------------------------------------------
+
+namespace Akka.Event;
```
Aaronontheweb: just use some MSBuild variables to not apply that global using to an Aspire project

examples/Simple/SimpleConsumer.AppHost/SimpleConsumer.AppHost.csproj
Arkatufus: Done

examples/Simple/SimpleConsumer.AppHost/Stub.cs
Arkatufus: Done

PR #500, opened 2025-08-12
Fix race condition in SubSourceLogic when consumer is aborted

### LINKED ISSUE #499: Race condition in RebalanceExtTests when consumer is aborted
## Description
There is a race condition in `RebalanceExtTests.FetchedRecordsMustNotBeLostUponAbort` that causes intermittent test failures in CI.

## Error Message
```
System.AggregateException : One or more errors occurred. (Failed: Timeout 00:00:03 while waiting for a message of type Akka.Streams.Kafka.Settings.TopicPartitionsRevoked )
---- Failed: Timeout 00:00:03 while waiting for a message of type Akka.Streams.Kafka.Settings.TopicPartitionsRevoked
```

## Root Cause
When a consumer is aborted with a kill switch while processing messages, it enters an infinite loop trying to recreate SubSources for partitions it no longer owns. This happens because:

1. The `HandleSubsourceCancelled` method in `SubSourceLogic.cs` uses a `ReEmit` strategy that immediately re-adds the partition to pending and tries to re-emit it
2. When the consumer is being shut down/aborted, this creates an infinite loop of SubSource creation attempts
3. The rapid SubSource recreation prevents proper partition revocation notifications from being sent

## Evidence from Logs
The logs show Consumer-1 repeatedly trying to start SubSource for partition [[1]] after being aborted (20+ attempts in rapid succession), each failing with the abort exception but immediately retrying due to the `ReEmit` strategy.

## Solution
The fix involves:
1. Checking if the stage is shutting down (`_shutdownTokenSource.IsCancellationRequested`) before re-emitting partitions
2. Checking if a partition is scheduled for revocation before re-emitting
3. Using `DoNothing` strategy instead of `ReEmit` when the outlet is closed

## Files Affected
- `src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs`
- `src/Akka.Streams.Kafka.Tests/Integration/RebalanceExtTests.cs`

## Related Information
- Build failure: https://github.com/akkadotnet/Akka.Streams.Kafka/runs/47928033695
- Previous race condition fix: #498

## Fix Status
A fix has been implemented and tested successfully. The test now passes consistently across multiple runs.

Aaronontheweb:
## Changes

When a consumer is aborted via kill switch, it could enter an infinite loop trying to recreate SubSources for partitions it no longer owns. This was caused by the ReEmit strategy immediately re-adding partitions to pending without checking if the stage is shutting down.

Changes:
- Check if stage is shutting down before re-emitting partitions
- Check if partition is scheduled for revocation before re-emitting
- Use DoNothing strategy when outlet is closed to prevent loops
- Add debug logging for better troubleshooting

This fixes the intermittent test failure in `RebalanceExtTests.FetchedRecordsMustNotBeLostUponAbort`

Fixes #499

## Checklist

* [x] Design discussion issue #499

### DIFF
--- a/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
+++ b/src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
@@ -198,6 +198,38 @@ internal class SubSourceLogic<K, V, TMessage> : TimerGraphStageLogic
         SetHandler(shape.Outlet, EmitSubSourcesForPendingPartitions, _ => PerformShutdown());
     }
 
+    private Func<IImmutableSet<TopicPartition>, IImmutableSet<TopicPartitionOffset>>? ConvertOffsetProvider(
+        Func<IImmutableSet<TopicPartition>, Task<IImmutableSet<TopicPartitionOffset>>> asyncProvider)
+    {
+        return partitions =>
+        {
+            try
+            {
+                // We need to block here because Confluent.Kafka's handler is synchronous
+                // This should be fast as the offset provider typically just returns stored offsets
+                var task = asyncProvider(partitions);
+                if (task.IsCompleted)
+                {
+                    return task.Result;
+                }
+                
+                // Use a reasonable timeout to avoid blocking forever
+                if (task.Wait(TimeSpan.FromSeconds(5)))
+                {
+                    return task.Result;
+                }
+                
+                Log.Warning("Offset provider timed out for partitions: {0}", string.Join(", ", partitions));
+                return ImmutableHashSet<TopicPartitionOffset>.Empty;
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Error in offset provider for partitions: {0}", string.Join(", ", partitions));
+                return ImmutableHashSet<TopicPartitionOffset>.Empty;
+            }
+        };
+    }
+    
     protected void ConfigureSubscription(Action<IImmutableSet<TopicPartition>> partitionsAssignedCb,
         Action<IImmutableSet<TopicPartitionOffset>> partitionsRevokedCb)
     {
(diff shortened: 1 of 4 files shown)


src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -488,6 +488,22 @@ private void HandleSubsourceCancelled((TopicPartition, ISubSourceCancellationStr
         _subSources = _subSources.Remove(topicPartition);
         _partitionsInStartup = _partitionsInStartup.Remove(topicPartition);
 
+        // If the stage is shutting down, don't re-emit partitions to avoid infinite loops
+        if (_shutdownTokenSource.IsCancellationRequested)
```
Aaronontheweb: Comment above says it all

src/Akka.Streams.Kafka/Stages/Consumers/Abstract/SubSourceLogic.cs
```diff
@@ -488,6 +488,22 @@ private void HandleSubsourceCancelled((TopicPartition, ISubSourceCancellationStr
...
+        {
+            if (Log.IsDebugEnabled)
+                Log.Debug("#{0} Stage is shutting down, not re-emitting partition {1}", _actorNumber, topicPartition);
+            return;
+        }
+
+        // Check if this partition is scheduled for revocation - if so, don't re-emit
+        if (_partitionsToRevoke.Contains(topicPartition))
```
Aaronontheweb, line 492: If we're not shutting down, but this partition is being revoked, don't re-emit it

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

